PR #2, opened 2018-08-16
Enh/pm_Set<int64_t>

kalmarek:
Things that work:

```julia
julia> using PolymakeWrap
polymake version 3.2
Copyright (c) 1997-2018
Ewgenij Gawrilow, Michael Joswig (TU Berlin)
http://www.polymake.org

julia> PolymakeWrap.pm_Set()
pm::Set<long, pm::operations::cmp>
{}

julia> PolymakeWrap.pm_Set([1,2,3])
pm::Set<long, pm::operations::cmp>
{1 2 3}

julia> PolymakeWrap.pm_Set([1,2,3,2])
pm::Set<long, pm::operations::cmp>
{1 2 3}

julia> PolymakeWrap.pm_Set([1,2,3,2,1])
pm::Set<long, pm::operations::cmp>
{1 2 3}

julia> arr = rand(-5:5, 10) 
10-element Array{Int64,1}:
 -5
  3
  0
 -5
 -3
 -3
 -1
  4
 -2
  4

julia> PolymakeWrap.pm_Set(arr)
pm::Set<long, pm::operations::cmp>
{-5 -3 -2 -1 0 3 4}

```
TODO: parametrize `pm_Set` by the type of elements

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -16,6 +16,10 @@ function convert(::Type{Polymake.pm_Integer},int::T) where T <: Union{Int128,Int
     return Polymake.pm_Integer(int)
 end
 
+function convert(::Type{Polymake.pm_Set}, A::Vector{T}) where T<:Signed
+    return Polymake.new_set_int64(A)
+end
+
 function convert(::Type{Rational{BigInt}},rat::Polymake.pm_Rational)
     num = Polymake.numerator(rat)
     denom = Polymake.denominator(rat)
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ deps/src/CMakeFiles/*
 deps/src/Makefile
 deps/src/cmake_install.cmake
 deps/src/libpolymake.so
+local
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -4,7 +4,7 @@ using CxxWrap
 
 if ! haskey(ENV, "POLYMAKE_CONFIG")
     const oldwdir = pwd()
-    const pkgdir = Pkg.dir("Polymake")
+    const pkgdir = Pkg.dir("PolymakeWrap")
     const nemodir = Pkg.dir("Nemo")
 
     wdir = "$pkgdir/deps"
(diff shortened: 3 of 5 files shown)


sebasguts:
Very nice, thanks a lot!

src/convert.jl
```diff
@@ -16,6 +16,10 @@ function convert(::Type{Polymake.pm_Integer},int::T) where T <: Union{Int128,Int
     return Polymake.pm_Integer(int)
 end
 
+function convert(::Type{Polymake.pm_Set}, A::Vector{T}) where T<:Signed
```
sebasguts: Is this just WIP? Otherwise, shouldn't the `T` be inherited from `Integer`?

sebasguts:
If you want to, I can merge this now. I would get rid of the merge commits before, though. I think the next step would be a conversion back from `Set<int>` to maybe an array (or a set data stucture, I do not care too much actually). I think this is even more important than having a parametrized set.

src/convert.jl
kalmarek: All of this is a very first-try effort ;-)

`Signed` is the type of all machine-style signed integers. I don't make any promises of the sets of arbitrary sized ints, at least for now. And it's a placeholder, until parametrized set lands anyway.

kalmarek:
I can work with on this branch, but on Monday I go for 2 weeks of holiday, so if You plan on making strides on the master it'd be better to merge it then (maybe on Sunday?)

So I'd rather not deal with mirroring the data structure of `pm::Set` as a `struct` in Julia for direct conversion. I'd rather to convert to build e.g. an array on the julia side and fill it with elements of `pm::Set`.

src/convert.jl
sebasguts: Okay, then it is perfect!

sebasguts:
Yeah, I think converting the set to an array is a very good idea. I think sticking to standard julia data structures is a good idea in general. But julia also has a `Set` data structure, so maybe convert to that one?

There is no need to have everything perfect for now. Since the comment I had was not valid anyway, I will merge it now, and then the conversion to the other side can be  new PR.

PR #3, opened 2018-09-03
Enh/parametrized set

kalmarek:


kalmarek:
this is complete* implementation of julia Sets on top of pm_Set; have a look at tests to see what is possible.

* one missing important bit: iteration. That requires creating an "iterable thing" which produces the next state/element of iterable. In this case that would require to define the pm::entire(pm::Set) iterator in CxxWrap Polymake module. We should probably do that anyway, but that's another task.

sebasguts:
Ah, I did a pretty bad mistake now tbh, because this should probably not have been squashed. Do you want me to revert the PR, so you can open it again and I merge it without squashing?

kalmarek:
Thanks;
I think it's fine, I'll just delete this branch and branch again from master.

sebasguts:
The thing is that your commits are lost then. You decide.

kalmarek:
that's fine, 
I was thinking about rewriting history of all of this, but squash makes it easier for me ;-) 

in the case I'm looking for a lonely commit I still have my branch

PR #5, opened 2018-09-26
The Big Rename

saschatimme:
This changes the repo name from `PolymakeWrap.jl` to `Polymake.jl` and the C++ wrapper from `Polymake` to `CxxPM` (shorter since we write this a lot). This should superseed #1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,9 @@ git:
 
 ## uncomment the following lines to override the default test script
 #script:
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("PolymakeWrap"); Pkg.test("PolymakeWrap"; coverage=true)'
+#  - julia -e 'Pkg.clone(pwd()); Pkg.build("Polymake"); Pkg.test("Polymake"; coverage=true)'
 after_success:
   # push coverage results to Coveralls
-  - julia -e 'cd(Pkg.dir("PolymakeWrap")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'cd(Pkg.dir("Polymake")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
   # push coverage results to Codecov
-  - julia -e 'cd(Pkg.dir("PolymakeWrap")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
+  - julia -e 'cd(Pkg.dir("Polymake")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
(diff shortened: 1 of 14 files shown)


saschatimme:
I think it would be better to drop this nested module structure all along. This is also what the CxxWrap guy does in [QML.jl](https://github.com/barche/QML.jl/blob/master/src/QML.jl).

sebasguts:
Using the nested module stuff provides a clear separation between C++ and Julia functions.

I would vote for leaving the C++ functions in the internal module, and only exposing those in the outer module that should be called by the user. Some of the C++-functions might not be save to call (note that we for example sometimes make assumptions about the input data layout and such) and should really not be exposed to the outer world, not even as `Polymake.foo`.

In fact, we will have even more internal modules soon, which can also not really be exposed (or at least not by now). I will create an internal module for each polymake application. This is necessary to mirror polymakes functionality to the julia side.

sebasguts:
Can you come up with any immediate benefit for dropping the separation?

saschatimme:
The one thing which is not clear to me, is how we will expose functions to the end user.
Do we write for each function we want to "export" a new Julia function which just delegates stuff?

Example
```julia
# currently
julia> Polymake.CxxPM.cube(4)

# Custom export

# 1) define
module Polymake

   export cube
   .....
   cube(args...) = CxxPM.cube(args...)
   .....
end
julia> Polymake.cube(4)
```

Assuming we do not have the nested structure we just can `export` all "safe" functions.

```julia
module Polymake

       using CxxWrap

       @wrapmodule(.....)

       export cube
end
```

And a user can also just call
```julia
julia> Polymake.cube(4)
```

sebasguts:
I would go for explicit definitions a la
```
const cube = PolymakeInternal.cube
```
This cube can then be exported. Is there something against keeping such a list?

I am not completely sure we should export functions at all, but I might have not yet understood it. Can I explicitly import/using a function that has not been exported, i.e., even with `cube` not being exported from the `Polymake` module, can I wrilte
```
import Polymake: cube
cube(4)
```
or does `cube` has to be exported to do so?

saschatimme:
> This cube can then be exported. Is there something against keeping such a list?

Keeping such a list feels somewhat redundant resp. we have another code generation we have to write (and this time on the Julia side) since I assume you don't want to do this by hand. I am not against the nested module structure, I just want to point the consequences.

> I am not completely sure we should export functions at all, but I might have not yet understood it. 

I think exporting functions serve two purposes in Julia. Since there are no public / private functions, you can see exported functions as the officially supported API. Another important feature is that if you do `using Polymake` and every exported function will be available without the need of a prefix.  (The `import Polymake: cube` works regardless of `export cube` or not). While this is not a best practice in package development, it is quite handy in the REPL.

Example
```julia
julia> module Foo
           cube(n) = "FooCube $n"
       end;

julia> module Bar
             export cube
             cube(n) = "BarCube $n"
       end;

julia> using .Foo

julia> cube(4)
ERROR: UndefVarError: cube not defined
Stacktrace:
 [1] top-level scope at none:0

julia> using .Bar

julia> cube(4)
"BarCube 4"
```

saschatimme:
Closed in favor of #26.

PR #12, opened 2018-10-07
Added first test version of parser.

sebasguts:


deps/parser/parser.jl
```diff
@@ -0,0 +1,184 @@
+using JSON
+
+typestring = ""
+undotted_typestring = typestring[1:length(typestring)-1]
```
kalmarek: what is typestring intended for? future namespacing?
If You need to get rid of the dot at the end use `strip(typestring, ".")`

deps/parser/parser.jl
```diff
@@ -0,0 +1,184 @@
...
+
+function create_argument_string_from_type( type_string::String, number::Int64 )
+    ## Special case for BigObject (we do not care 'bout the type here)
+    if startswith(type_string,"BigObject")
+        return "arg" * string(number) * "::" * type_translate_list["BigObject"]
+    end
+    ## We only translate the known cases
+    if !haskey(type_translate_list,type_string)
+        println("Unknown type: " * type_string)
+        return false
```
kalmarek: here we could just throw an 
```julia
struct UnknownPolymakeType <: Exception
    msg::String
    UnknownPolymakeType(type_string) = new("Encountered an unknown polymake type: $type_string")
end
```

deps/parser/parser.jl
```diff
@@ -0,0 +1,184 @@
...
+        ## Fixme: good?
+        min_argument_number = min_argument_number + 1
+        max_argument_number = max_argument_number + 1
+    else
+        argument_list_header = Array{String,1}()
+    end
+    argument_list = Array{String,1}()
+
+    for i in 1:max_argument_number
+        push!(argument_list_header,create_argument_string_from_type(arguments[i],i))
```
kalmarek:
```julia
try
    push!(argument_list_header,create_argument_string_from_type(arguments[i],i))
catch ex
    if ex isa UnknownPolymakeType
         ...
    else
         rethrow(ex)
    end
end

File: deps/parser/parser.jl [hunk 4]
```diff
@@ -0,0 +1,184 @@
...
+                    return false
+                else
+                    println("Ignoring " * string(max_argument_number-i) * " arguments of function " * name)
+                    max_argument_number = i - 1
+                end
+            end
+        end
+    end
+
+    return_string = ""
```
kalmarek: this is probably the place to separate json parsing from code generation

deps/parser/parser.jl
```diff
@@ -0,0 +1,184 @@
...
+module $app_name
+
+import ..pm_Integer, ..pm_Rational, ..pm_Matrix, ..pm_Vector, ..pm_Set,
+       ..pm_perl_Object, ..pm_perl_OptionSet, ..pm_perl_PropertyValue,
+       ..Polymake.application, ..Polymake.call_function, ..Polymake.call_method
+
+"""
+    for current_function in parsed_dict["functions"]
+        return_value = parse_function_definition(current_function,app_name)
+        if return_value == false
```
kalmarek: or even catch that exception again here

deps/parser/parser.jl
```diff
@@ -0,0 +1,184 @@
...
+        push!(argument_list_header,create_argument_string_from_type(arguments[i],i))
+        push!(argument_list,"arg$i")
+    end
+
+    if false in argument_list_header
+        for i in 1:length(argument_list_header)
+            if argument_list_header[i] == false
+                if i <= min_argument_number
+                    println("Cannot parse function " * name * " due to unparsable argument")
+                    return false
```
kalmarek: a pretty terrible way of signaling a failure ;-)

deps/parser/parser.jl
sebasguts: This is not supposed to throw an exception, this is just supposed to warn me.

deps/parser/parser.jl
sebasguts: Yep. We have not yet resolved the internal module. It was previously used in different places, I guess we can actually remove it now.

deps/parser/parser.jl
```diff
@@ -0,0 +1,184 @@
...
+    for current_function in parsed_dict["functions"]
+        return_value = parse_function_definition(current_function,app_name)
+        if return_value == false
+            println("Unable to parse " * string(current_function))
+        else
+            return_string = return_string * return_value
+        end
+    end
+    return_string = return_string * "\n\nend\nexport $app_name\n"
+    open(abspath(outputfileposition * "/" * app_name * ".jl" ),"w") do outputfile
```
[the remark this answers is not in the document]
sebasguts: Does `joinpath` also work on windows?

deps/parser/parser.jl
sebasguts: See above

deps/parser/parser.jl
sebasguts: Please elaborate?

kalmarek:
Did You think about separating the flow of `parse_function_definition` for methods and functions?

I'd probably made a types `PolymakeFunction`/ `PolymakeMethod` and dispatch code generation on them

but we can work on this later, if You want to have this merged fast

deps/parser/parser.jl
sebasguts: As said above, I wanted to have this information printed, not really providing any exception handling. If Julia has a standard way to print warnings, we could use that one instead. I definitely do not want an error here.

sebasguts:
I thought about separating the flow first, but the changes were too minimal. I wanted to avoid unnecessary code duplication.

deps/parser/parser.jl
kalmarek: then `@warn(...)`;
but You check for `falses` above, so this may as well throw with handling doing the warning

deps/parser/parser.jl
kalmarek: if You catch you can just print a warning and continue as if nothing has happened

deps/parser/parser.jl
kalmarek: below we mostly fill the return_string with all the data we pulled (and transformed) from json;
and the name of the function reads `parse...`, whereas it does both parsing and code gen

sebasguts: K.

kalmarek:
@sebasguts could You also push some json with methods? the one in this pull has only functions

deps/parser/parser.jl
```diff
@@ -0,0 +1,184 @@
...
+end
+
+"""
+end
+
+function julia_method_string(julia_name::String, polymake_name::String, app_name::String, julia_args::String, call_args::String, parameter_string::String )
+    return """
+function $julia_name( $julia_args )
+    $(typestring)application( "$app_name" )
+    return $(typestring)call_method( "$polymake_name$parameter_string", dispatch_obj, Array{Any,1}([ $call_args ]) )
```
saschatimme: Is this `dispatch_obj` really always the same? Will this implicitly be in the `julia_args` list?

saschatimme:
For me it is also fine if we merge the parser for now (such that Sebastian can do other stuff) and I prettify everything afterwards. Since Sebastian probably has more important stuff to do than producing pretty Julia code. That's something where I can help out.

kalmarek:
I am fine with this as well, @sebasguts just push more examples.

now, where do I approve this thing?

deps/parser/parser.jl
saschatimme: I see this is implictly provided

sebasguts: Yep, it is always the same. The methods are object methods.

sebasguts:
I will do the example stuff and some minor refactoring now, just because I also need to learn how to produce pretty Julia code. Full blown exception handling is a bit overkill here, but I probably have just worked with too many languages that do not have any, so I need to get used to it again. I will push my changes tonight, then you can have another look and we merge this stuff tomorrow.

I am not completely sure this will be the finished thing, I still need to wait for some more JSON to see if I found all corner cases (Spoiler alert: Probably not). Afterwards, once the JSON is there, we can mock up some test cases.

deps/parser/parser.jl
sebasguts: I decided to not do this at the moment. The name of the function in the current state is probably debatable, but currently there is really no use in ripping this function apart into two.

Granted, two would be a bit more clean, but you end up in having to parse all the extracted information from the first part to the second part. Either using keyword arguments or a dict, since everything else would be very messy.

But since both parts are resonably small, I would leave them as they are.

sebasguts:
So, I have added the exceptions, plus minor comments from @kalmarek (Using `strip` and `joinpath`, mostly). I also renamed the large function into `parse_definition`. Tomorrow I will make up more JSON to test more corner cases, and specially the method case. I would suggest to not merge this PR yet. You can comment and review, but I would suggest merging it once it is finished.

kalmarek:
following Saschas comment I'm for merging it and prettifying later

sebasguts:
It is not really about prettifying later, but about fixing it up with more examples and some more changes from the polymake side. As said, I would rather continue to work on this branch until it is really somehow "finished", at least beta or something.

kalmarek:
could You share a json with "method"?

deps/parser/parser.jl
```diff
@@ -0,0 +1,203 @@
...
+    is_method = haskey(method_dict,"method_of")
+    ## for now, we use the same name for Julia and Polymake
+    julia_name = name
+    polymake_name = name
+
+    ## Check for option set
+    has_option_set = last(arguments) == "OptionSet"
+
+    ## Make type parameters
+    param_list_header = Array{String,1}()
```
kalmarek: this could be made into a separate function:
```julia
function params(n::Int)
    if n == 0
        return Array{String,1}(), ""
    end

    param_list_header = Array{String,1}()
    param_list = Array{String,1}()

    for i in 1:n
        push!(param_list_header,"param$i::String")
        push!(param_list,"\$param$i")
    end

    param_string = "<" * join(param_list,",") * ">"

    return param_list_header, param_string
end
```

deps/parser/parser.jl
```diff
@@ -0,0 +1,203 @@
...
+        push!(param_list,"\$param$i")
+    end
+    if length(param_list) >= 1
+        parameter_string = "<" * join(param_list,",") * ">"
+    else
+        parameter_string = ""
+    end
+
+    ## Compute the argument range
+    max_argument_number = length(arguments) - (has_option_set ? 1 : 0)
```
kalmarek:
```julia
max_args(len::Int, has_option::Bool) = len - (has_option ? 1 : 0)
```
then
```
if is_method
    min_arg_number = mandatory + 1
    min_arg_number = max_args(length(args), has_option_set) + 1
    argument_list_header = Array{String,1}(["dispatch_obj::$(typestring)pm_perl_Object"])
else # is_function
    min_arg_number = mandatory
    min_arg_number = max_args(length(args), has_option_set)
    argument_list_header = Array{String,1}()
end
```

deps/parser/parser.jl
```diff
@@ -0,0 +1,203 @@
...
+        argument_list_header = Array{String,1}(["dispatch_obj::$(typestring)pm_perl_Object"])
+        ## Fixme: good?
+        min_argument_number = min_argument_number + 1
+        max_argument_number = max_argument_number + 1
+    else
+        argument_list_header = Array{String,1}()
+    end
+    argument_list = Array{String,1}()
+
+    for i in 1:max_argument_number
```
kalmarek: I'd be best to make it into a `function argument_list(arguments, max_argument_number)` which
`return argument_list_header, argument_list`

deps/parser/parser.jl
```diff
@@ -0,0 +1,203 @@
...
+            push!(argument_list_header,create_argument_string_from_type(arguments[i],i))
+        catch exception
+            if exception isa UnknownPolymakeType
+                @warn(exception.msg)
+                break
+            else
+                rethrow(exception)
+            end
+        end
+        push!(argument_list,"arg$i")
```
kalmarek: this is of little importance here, but repeated pushing is not the best thing here; comprehension ala
`argument_list = Vector{String}(["arg$i" for i in 1:max_argument_number])`
is better

kalmarek:
@sebasguts also the julian way to handle this would be to create separate types:
```julia
abstract type PMCallable end

struct PMFunction <: PMCallable
    app::String
    name::String
    args::Vector{String}
    mandatory::Int
    type_params::Int
end

struct PMMethod <: PMCallable
    app::String
    name::String
    args::Vector{String}
    mandatory::Int
    type_params::Int
end

function PMCallable(app_name::String, method_dict::Dict)
    if haskey(callable_dict, "method_of")
        return PMMethod(app_name, callable_dict)
    else
        return PMFunction(app_name, callable_dict)
    end
end
```
and then instead of ifing dispatch function on the type:

```julia
julia_name(c::PMCallable) = c.name
polymake_name(c::PMCallable) = c.name

has_option_set(c::PMCallable) = last(c.args) == "OptionSet"

max_args(len::Int, has_option::Bool) = len - (has_option ? 1 : 0)

max_arg_number(f::PMFunction) = max_args(length(f.args), has_option_set(f))
min_arg_number(f::PMFunction) = c.mandatory

min_arg_number(m::PMMethod) = c.mandatory + 1
min_arg_number(m::PMMethod) = max_args(length(c.args), has_option_set(c)) + 1
```
And so on.
We (Sascha and I) are more used to writing such code, so we could PR that for You, if You wish

deps/parser/parser.jl
saschatimme: You can even drop the `Vector{String}` constructor.

deps/parser/parser.jl
```diff
@@ -30,26 +30,39 @@ type_translate_list = Dict(
...
+function $julia_name( $julia_args, ::Val{false} )
     $(typestring)application( "$app_name" )
     return $(typestring)call_function( "$polymake_name$parameter_string", Array{Any,1}([ $call_args ]) )
 end
 
+function $julia_name( $julia_args )
+    $(typestring)application( "$app_name" )
+    return_value = $(typestring)call_function( "$polymake_name$parameter_string", Array{Any,1}([ $call_args ]) )
+    return convert_from_property_value(return_value)
+end
```
saschatimme: You could also have just one function definition, the compiler figures that out.

```julia
function $julia_name( $julia_args , ::Val{Convert}=Val(true))
     $(typestring)application( "$app_name" )
     return_value = $(typestring)call_function( "$polymake_name$parameter_string", Array{Any,1}([ $call_args ]) )
     if Convert
         return convert_from_property_value(return_value)
    else
        return return_value
    end
end
```

sebasguts: I see this works. I will do this.

sebasguts:
Closing this for a new one

PR #16, opened 2018-10-08
Enh/simplify sets

kalmarek:
* make `pm_Set <: AbstractSet`
* reduce the int types on the C++ side to minimum
* move all the convenience to julia side
* reduce julia side to constructors and `AbstractSet` interface
* add the adjusted julia-Sets tests for `pm_Set`


src/sets.jl
```diff
@@ -13,12 +13,18 @@ for (T, f) in [
...
 pm_Set{T}(s::Set) where T = pm_Set{T}(collect(s))
 
-pm_Set(::Type{T}) where T = pm_Set{T}()
+pm_Set{S}(n::T) where {S,T <: Integer} = Polymake.scalar2set(S(n))
+
+pm_Set{T}(itr) where T = union!(pm_Set{T}(), itr)
+
+function pm_Set(itr)
+    # use IteratorEltype(itr) trait?
+    T = typeof(first(itr))
```
saschatimme: This can lead to problems if the iterator is stateful. We could loose the first element then.

saschatimme: I think it would be fine to rely on `eltype(itr)`

kalmarek: yeah, I had this with `eltype` before, but
```julia
julia> eltype(1:5)
Int64

julia> eltype(x for x in 1:5)
Any
```
julia Sets do
```julia
Set(itr) = _Set(itr, IteratorEltype(itr))

_Set(itr, ::HasEltype) = Set{eltype(itr)}(itr)

function _Set(itr, ::EltypeUnknown)
    T = @default_eltype(itr)
    (isconcretetype(T) || T === Union{}) || return grow_to!(Set{T}(), itr)
    return Set{T}(itr)
end
```
is `T = @default_eltype(itr)` good enough?

saschatimme: No idea what `T = @default_eltype(itr)` is doing but probably yes.

kalmarek: `@edit Base.@default_eltype(x for x in 1:5)`

kalmarek: well, not good enough, `@default_eltype` may return unions, etc:
```
Construction: Error During Test at /home/kalmar/.julia/dev/PolymakeWrap/test/sets.jl:271
  Got exception outside of a @test
  MethodError: no method matching PolymakeWrap.Polymake.pm_Set{Union{Int32, Int64}}()
  Closest candidates are:
    PolymakeWrap.Polymake.pm_Set{Union{Int32, Int64}}(!Matched::Array{T,1} where T) where T at /home/kalmar/.julia/dev/PolymakeWrap/src/sets.jl:16
...
```

saschatimme: For new also can go the `Set(collect(iter))` way instead. Not optimal, but whatever.

saschatimme: At leas if `eltype(T)` is not a concrete type.

saschatimme:
@sebasguts I am not authorized to merge this PR. Can you merge it or give me the right permissions?

sebasguts:
That is strange, you should be able to merge. I will check.

sebasguts:
I have a minor comment on this. You are using the set typename `Set`. While this is currently okay, we might at some point decide to use the namespace `pm` completely (which was already suggested by @benlorenz). While this might not cause trouble (I would have to check), it will nevertheless be confusing. Could you name it differently? Maybe `pm_Set` or `Set_class`?

PR #18, opened 2018-10-09
Better Integers

saschatimme:
This is not yet finished, but I wanted to put this here just that you know what is happening.

### DIFF
--- a/deps/src/polymake_integers.cpp
+++ b/deps/src/polymake_integers.cpp
@@ -32,7 +32,7 @@ void polymake_module_add_integer(jlcxx::Module& polymake){
     .method("<=", [](int32_t a, pm::Integer& b){ return a <= b; })
 
     .method("show_small_obj", [](pm::Integer& i){
-      return show_small_object<pm::Integer>(i);
+      return show_small_object<pm::Integer>(i, false);
     })
     // the symmetric definitions are on the julia side
     .method("+",[](pm::Integer &a, pm::Integer &b){ return a + b; })
--- a/deps/src/polymake_integers.cpp
+++ b/deps/src/polymake_integers.cpp
@@ -38,16 +38,32 @@ void polymake_module_add_integer(jlcxx::Module& polymake){
     .method("+",[](pm::Integer &a, pm::Integer &b){ return a + b; })
     .method("+",[](pm::Integer &a, int64_t b){ return a + static_cast<long>(b); })
     .method("+",[](pm::Integer &a, int32_t b){ return a + b; })
-    // the symmetric definitions are on the julia side
+    .method("+",[](int64_t a, pm::Integer &b){ return static_cast<long>(a) + b; })
+    .method("+",[](int32_t a, pm::Integer &b){ return a + b; })
+
     .method("*",[](pm::Integer &a, pm::Integer &b){ return a * b; })
     .method("*",[](pm::Integer &a, int64_t b){ return a * static_cast<long>(b); })
     .method("*",[](pm::Integer &a, int32_t b){ return a * b; })
+    .method("*",[](int64_t a, pm::Integer &b){ return static_cast<long>(a) * b; })
+    .method("*",[](int32_t a, pm::Integer &b){ return a * b; })
 
     .method("-",[](pm::Integer &a, pm::Integer &b){ return a - b; })
     .method("-",[](pm::Integer &a, int64_t b){ return a - static_cast<long>(b); })
     .method("-",[](pm::Integer &a, int32_t b){ return a - b; })
     .method("-",[](int64_t a, pm::Integer &b){ return static_cast<long>(a) - b; })
-    .method("-",[](int32_t a, pm::Integer &b){ return a - b; });
+    .method("-",[](int32_t a, pm::Integer &b){ return a - b; })
+
+    .method("div",[](pm::Integer &a, pm::Integer &b){ return a / b; })
+    .method("div",[](pm::Integer &a, int64_t b){ return a / static_cast<long>(b); })
+    .method("div",[](pm::Integer &a, int32_t b){ return a / b; })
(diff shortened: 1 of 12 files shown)


kalmarek:
this will conflict with `enh/simplify_sets` as I've already put tests for sets in a separate file

saschatimme:
I will just rebase then

deps/src/polymake_integers.cpp
```diff
@@ -6,19 +6,31 @@
 
 #include "polymake_integers.h"
 
-pm::Integer new_integer_from_bigint(jl_value_t* integer){
+pm::Integer cast_bigint_to_integer(jl_value_t* integer){
```
sebasguts: Please note that this is not a cast, but indeed copies the integer. Everything else is unsafe.

The copy is implicit in the way CxxWrap works: by returning the dereferenced `p` CxxWrap actually needs to create a copy of `p`, and therefore the integer.

saschatimme: Oh good to know. What is about the definition on the Julia side? Does this copy?

saschatimme:
```julia
"""
    bigint!(int::Polymake.pm_IntegerAllocated)

Cast a `Polymake.pm_IntegerAllocated` to a `BigInt`. This *does not* create a copy.
"""
function bigint!(int::Polymake.pm_IntegerAllocated)
    unsafe_load(reinterpret(Ptr{BigInt},int.cpp_object))
end
```

sebasguts: No, it does not.

`unsafe_load` in Julia just gives you an object ref to whatever the pointer points to. It would be safer to actually create a copy there.

sebasguts:
Argh, of course my try to resolve this caused an error. I will fix that commit

deps/src/polymake_integers.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+void polymake_module_add_integer(jlcxx::Module& polymake){
+  polymake.add_type<pm::Integer>("pm_Integer", jlcxx::julia_type("Integer", "Base"))
+    .constructor<int32_t>()
+    .constructor<int64_t>()
+    // the symmetric definitions are on the julia side
+    .method("+",[](pm::Integer &a, pm::Integer &b){ return a + b; })
+    .method("+",[](pm::Integer &a, int64_t b){ return a + static_cast<long>(b); })
+    .method("+",[](pm::Integer &a, int32_t b){ return a + b; })
+    // the symmetric definitions are on the julia side
```
sebasguts: While I am technically fine with this, as they seem to be magically there, there are two potential problems here:
1. This adds another indirection in function calls. While Julia should technically be able to resolve this, we should look at the assembly if it really gets rid of the function call. Otherwise simply add the symmetric devs here, since this should cause stupid performance decreases otherwise.
2. Should we have a special addition for BigInt? Currently, Adding a `pm_Integer` and a `BigInt` creates a new `pm_Integer` from the `BigInt`. While `BigInt`s are not that popular anyway, this could also potentially cause strange performance penalties.

src/integers.jl
```diff
@@ -0,0 +1,50 @@
...
+Polymake.pm_Integer(int::BigInt) = Polymake.cast_bigint_to_pm_integer!(copy(int))
+Polymake.pm_Integer(int::Integer) = Polymake.pm_Integer(convert(BigInt, int))
+
+"""
+    bigint!(int::Polymake.pm_IntegerAllocated)
+
+Cast a `Polymake.pm_IntegerAllocated` to a `BigInt`. This *does not* create a copy.
+"""
+function bigint!(int::Polymake.pm_IntegerAllocated)
+    unsafe_load(reinterpret(Ptr{BigInt},int.cpp_object))
```
sebasguts: I am not sure we should really do this. It is not an inplace bigint, it returns a new object that shares memory with the old one. So doing
```
x = bigint!(x)
```
or
```
bigint!(pm_Integer(0))
```
will (not only can, but will!) cause a segfault.A better name would be `unsafe_convert_to_bigint`, which makes it a bit clearer that it should not be used with values that get potentially destroyed too early.

I am honestly not even sure the way you use it below is safe, I would have to investigate a bit.

src/PolymakeWrap.jl
```diff
@@ -1,5 +1,8 @@
 module PolymakeWrap
 
+# we needs this here to resolve some type ambiguities
+import CxxWrap: SmartPointer
```
sebasguts: Why do you need the SmartPointers?

saschatimme:
@sebasguts In light of your concerns, I just resorted to always creating copies for the conversion between `pm_Integer` and `BigInt`. Since this is a performance improvement, we can do this at another point in time if necessary.

I think this can be now reviewed. I restricted the capabilities to the most basic one, but I think this should be sufficient for now.

deps/src/polymake_integers.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    .method("rem",[](pm::Integer &a, pm::Integer &b){ return a % b; })
+    .method("rem",[](pm::Integer &a, int64_t b){ return a % static_cast<long>(b); })
+    .method("rem",[](pm::Integer &a, int32_t b){ return a % b; })
+    .method("rem",[](int64_t a, pm::Integer &b){ return static_cast<long>(a) % b; })
+    .method("rem",[](int32_t a, pm::Integer &b){ return a % b; });
+
+  polymake.method("new_pm_Integer_from_bigint", new_integer_from_bigint);
+
+}
```
kalmarek: don't we need unary `-` as well?

saschatimme: Good catch

saschatimme:
@kalmarek Added unary minus

PR #20, opened 2018-10-16
Enh/general cleanup

kalmarek:
constructors and basic functionality for pm_Vector, pm_Matrix; e.g.
```julia
    pmV = PolymakeWrap.pm_Vector
    pmI = PolymakeWrap.pm_Integer
    v = [1,2,3]

    pm_v = pmV{pmI}(3)
    @test pmV(v) == v
    for (e1, e2) in zip(pmV, v)
        @test e1 == e2
    end
    pm_v .= v
    @test pm_v == v
    @test pmV((4*v)//4) == v
```
have a further look at tests to see what is possible;
Notes:
* tests run even with 'include("converts.jl")` commented
* conversion TO julia will require some converts
* iteration works out of the box
* no arithmetic is present, but for everything should work once we add arithmetic for elements (except multiplication of `pm_Matrix` of course)

### DIFF
--- a/src/PolymakeWrap.jl
+++ b/src/PolymakeWrap.jl
@@ -95,10 +95,11 @@ include("functions.jl")
 include("convert.jl")
 include("sets.jl")
 # to be moved to Vectors/Matrices
+
 pm_Integer(b::BigInt) = new_pm_Integer(b)
 pm_Rational(num::T, den::T) where T<:Integer = 
     pm_Rational(pm_Integer(num), pm_Integer(den))
-pm_Rational(r::R) where R<:Rational = pm_Rational(numerator(r), denominator(r))
+pm_Rational(r::Rational) = pm_Rational(numerator(r), denominator(r))
 Base.one(i::T) where T <:pm_Integer = pm_Integer(1)
 pm_Rational(int::Integer) = pm_Rational(int, one(int))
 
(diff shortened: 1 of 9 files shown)


src/PolymakeWrap.jl
```diff
@@ -96,17 +96,64 @@ include("convert.jl")
 include("sets.jl")
 # to be moved to Vectors/Matrices
 pm_Integer(b::BigInt) = new_pm_Integer(b)
-pm_Rational(num::BigInt, den::BigInt) = pm_Rational(pm_Integer(num), pm_Integer(den))
+pm_Rational(num::T, den::T) where T<:Integer = 
+    pm_Rational(pm_Integer(num), pm_Integer(den))
+pm_Rational(r::R) where R<:Rational = pm_Rational(numerator(r), denominator(r))
```
saschatimme:
```suggestion
pm_Rational(r::Rational) = pm_Rational(numerator(r), denominator(r))
```
is a little but shorter and avoids the type parameter. (Obviously this is nitpicking)

src/PolymakeWrap.jl
```diff
@@ -96,17 +96,64 @@ include("convert.jl")
...
-    return Polymake._setindex!(V, T(val), n)
+    Polymake._setindex!(V, T(val), Int(n))
+    return V
+end
+
+function Base.getindex(M::pmM, i::Integer, j::Integer) where pmM <: pm_Matrix 
+    1 <= i <= Polymake.rows(M) || throw(BoundsError(M, [i,j]))
+    1 <= j <= Polymake.cols(M) || throw(BoundsError(M, [i,j]))
+    return Polymake._getindex(M, Int(i), Int(j))
 end
```
saschatimme: The nice way would be to allow other to avoid the bounds checks if they want. This can be accomplished as follows:
```julia
Base.@propagate_inbounds function Base.getindex(M::pm_Matrix , i::Integer, j::Integer)
     @boundscheck 1 <= i <= Polymake.rows(M) || throw(BoundsError(M, [i,j]))
     @boundscheck 1 <= j <= Polymake.cols(M) || throw(BoundsError(M, [i,j]))
     return Polymake._getindex(M, Int(i), Int(j))
 end
```

src/PolymakeWrap.jl
```diff
@@ -96,17 +96,64 @@ include("convert.jl")
...
+
+function Base.setindex!(M::pmM, val, i::Integer, j::Integer) where {T, pmM <: pm_Matrix{T}}
+    1 <= i <= Polymake.rows(M) || throw(BoundsError(M, [i,j]))
+    1 <= j <= Polymake.cols(M) || throw(BoundsError(M, [i,j]))
+    Polymake._setindex!(M, T(val), Int(i), Int(j))
+    return M
 end
```
saschatimme: Same as for getindex

saschatimme:
@kalmarek I assume it would make sense to rebase my Integer related changes onto this (and to get this merged quickly), do you agree?

kalmarek:
absolutely, I'll fix those;
@sebasguts are You happy with this?

probably one needs still to perfect iterator protocol (e.g. issues with eltype)? but this can wait

saschatimme:
@sebasguts can you merge this if you are happy ( I am still not able to merge branches)

sebasguts:
@saschatimme I have set your permissions to write, I would have to check again.

@kalmarek Why is the printing so strange:
```
julia> PolymakeWrap.pm_Matrix{PolymakeWrap.pm_Integer}([ 1 0 0 ; 1 2 0 ; 1 0 2 ])
3×3 PolymakeWrap.Polymake.pm_MatrixAllocated{PolymakeWrap.Polymake.pm_Integer}:
 pm::Integer
1  pm::Integer
0  pm::Integer
0
 pm::Integer
1  pm::Integer
2  pm::Integer
0
 pm::Integer
1  pm::Integer
0  pm::Integer
2
```

saschatimme:
That's what I see if it helps:
screen shot 2018-10-17 at 10 46 16

kalmarek:
because it's using a standard printing of Array (e.g. with ellipsis if an Array is too long);
if You want a short version try:
```
m = PolymakeWrap.pm_Matrix{PolymakeWrap.pm_Integer}([ 1 0 0 ; 1 2 0 ; 1 0 2 ])
[m]
```
then it prints m as an element using a compact representation

This is due to which method of `show` we overload

saschatimme:
Can't we use the Polymake printing instead?

kalmarek:
This is due to which method of `show` we overload:
REPL uses `display` which calls `show(io, "text/plain", obj)`;
`print` calls `show(io, obj)` which we overload
`print(m)` uses exactly polymake string

sebasguts:
@saschatimme will investigate

@kalmarek So, do I understand this correctly, that this comes from the fact that we use a generic array printing method? I would suggest overloading the printing method for matrix with the original polymake output. While writing `[m]` looks like what I want, it is a bit too much to ask for.

saschatimme:
@kalmarek Yeah just overload also "text/plain". This is all a little bit annoying, but I think at some point we have found all MIME types we need.

kalmarek:
@saschatimme yeah I knew that, but it was a deliberate choice I made; changing that

sebasguts:
@kalmarek Yep, that looks good.

saschatimme:
But the source issue is probably that our `pm_Integer`s print so weird. Should we just print the number and omit the `pm::Integer` part?

sebasguts:
Yeah, that sounds reasonable.

kalmarek:
that is the string repr that Polymake gives us

sebasguts:
Since with @saschatimme 's fixes for the integers they should seamlessly integrate with the Julia integers.

saschatimme:
Let's do this: For now overload "text/plain". And if the integer stuff is merged we revisit this. Okay?

PR #21, opened 2018-10-17
Polytope constructor

sebasguts:


src/convert.jl
```diff
@@ -42,7 +42,7 @@ function convert(::Type{Polymake.pm_Matrix{Polymake.pm_Integer}}, matrix::Array{
     for i in 1:rows
         for j in 1:cols
             converted = convert(Polymake.pm_Integer,matrix[i,j])
-            Polymake.set_entry(pm_matrix, i-1, j-1, converted )
```
kalmarek: You don't need to convert, it's handled in `setindex!`

test/runtests.jl
```diff
@@ -201,4 +201,5 @@ end
     end
 
     include("sets.jl")
+    include("perlobj.jl")
```
saschatimme: Did you forget to check that file in?

sebasguts: Ah, screw me. Thanks for noticing.

sebasguts:
This is now contained in #23

PR #26, opened 2018-10-18
The Big Rename - Second try

saschatimme:
Superseeds #5. This drops the nested module structure and renames PolymakeWrap -> Polymake.

In order to test this locally you have to remove PolymakeWrap `pkg> rm PolymakeWrap`, rename the folder to Polymake and add it again. Also you probably need to delete the cmake files in `deps/src` before building again.


README.md
```diff
@@ -1,7 +1,7 @@
-# PolymakeWrap
+# Polymake
 
-[![Build Status](https://travis-ci.org/sebasguts/PolymakeWrap.jl.svg?branch=master)](https://travis-ci.org/sebasguts/PolymakeWrap.jl)
+[![Build Status](https://travis-ci.org/sebasguts/Polymake.jl.svg?branch=master)](https://travis-ci.org/sebasguts/Polymake.jl)
```
sebasguts: Shouldn't this point to `oscar-system`?

saschatimme: That's true, but we also didn't setup CI so far.

sebasguts: True. Which will be another delicate thing to do.

PR #27, opened 2018-10-18
DO NOT MERGE - Methods

sebasguts:


sebasguts:
Okay, we can just get rid of this one. I am not sure what I tried to solve here, but it seems unnecessary.

sebasguts:
I added `DIM` to #34 . So with #34, you can now do
`polytope.DIM(c)`

PR #28, opened 2018-10-20
Enh/arrays2

kalmarek:
everything is fixed right now

### DIFF
--- a/deps/src/polymake.cpp
+++ b/deps/src/polymake.cpp
@@ -74,8 +74,8 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
     return to_SmallObject<pm::Matrix<pm::Rational>>(pv);
   });
     
-  POLYMAKE_INSERT_TYPE_IN_MAP_SINGLE_TEMPLATE(pm_Matrix,pm_Integer);
-  POLYMAKE_INSERT_TYPE_IN_MAP_SINGLE_TEMPLATE(pm_Matrix,pm_Rational);
+  POLYMAKE_INSERT_TYPE_IN_MAP(pm_Matrix_pm_Integer);
+  POLYMAKE_INSERT_TYPE_IN_MAP(pm_Matrix_pm_Rational);
 
   polymake.add_type<jlcxx::Parametric<jlcxx::TypeVar<1>>>("pm_Vector", jlcxx::julia_type("AbstractVector", "Base"))
     .apply<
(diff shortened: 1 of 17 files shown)


saschatimme:
@kalmarek I am not sure I completely understand, is it sufficient if we talk on Friday about this?

kalmarek:
@saschatimme sure it can

kalmarek:
this is ready (save @saschatimme casts to `long`), 
there is a bunch of tests and a reasonably flexible/convenient interface;

Please have a look

deps/src/polymake.cpp
```diff
@@ -95,32 +104,43 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
...
+    PropertyValueHelper ph(p); 
+    return ph.get_typename();
+  });
 
   polymake_module_add_set(polymake);
   POLYMAKE_INSERT_TYPE_IN_MAP(pm_Set_Int64);
   POLYMAKE_INSERT_TYPE_IN_MAP(pm_Set_Int32);
+  
+  polymake_module_add_array(polymake);
+  POLYMAKE_INSERT_TYPE_IN_MAP(pm_Array_Int32);
```
sebasguts: I am not sure what should be the best practice, actually. If you include `polymake_caller.h` in the `polymake_array.cpp`, you could also call this macro in the polymake array file. Which version do you prefer?

deps/src/polymake_functions.h
```diff
@@ -12,22 +12,12 @@ polymake::perl::Object call_func_2args(std::string, int, int);
 
 pm::perl::Object to_perl_object(pm::perl::PropertyValue);
 
-bool to_bool(pm::perl::PropertyValue v);
```
sebasguts: I think removing those from the header is fine.

deps/src/polymake_arrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+      typedef typename decltype(wrapped)::type WrappedT;
+      typedef typename decltype(wrapped)::type::value_type elemType;
+      
+      wrapped.template constructor<int32_t>();
+      wrapped.template constructor<int32_t, elemType>();
+      wrapped.template constructor<int64_t>();
+      wrapped.template constructor<int64_t, elemType>();
+      
+      wrapped.method("_getindex", [](WrappedT& A, int64_t n){
+        return elemType(A[n-1]);
```
saschatimme:
```suggestion
        return elemType(A[static_cast<long>(n) - 1]);
```

deps/src/polymake_arrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+      });
+      wrapped.method("_setindex!",[](WrappedT& A, elemType val, int64_t n){
+          A[n-1] = val;
```
saschatimme:
```suggestion
          A[static_cast<long>(n) - -1] = val;
```

deps/src/polymake_arrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+      });
+      wrapped.method("length", &WrappedT::size);
+      wrapped.method("resize!", [](WrappedT& A, int64_t newsz){
+        A.resize(newsz);
```
saschatimme:
```suggestion
        A.resize(static_cast<long>(newsz));
```

deps/src/polymake.cpp
kalmarek: ok, should we always call `POLYMKE_INSERT_TYPE_IN_MAP` inside the appropriate cpp files?

sebasguts: I am not sure. It would probably be make more sense there, but it would also be more distributed and therefore easier to miss.

kalmarek: Maybe it's best to have it where the type is defined? @saschatimme what do You think?

PR #33, opened 2018-10-24
Use all CPU threads for compiling

sebasguts:


sebasguts:
I actually wanted to redo the whole build system.

But I think it is useful by itself, so I changed the title. You are free to merge it :).

kalmarek:
actually this makes the compilation slower on my machine (because of throttling), can we make it `-j$(div(Sys.CPU_THREADS,2))`?

PR #34, opened 2018-10-24
WIP: More temporary method defs

sebasguts:
* `AMBIENT_DIM`


saschatimme:
Can we merge this, or should we forget about this?

sebasguts:
Nono, this is pretty important. I will finish it today.

sebasguts:
@saschatimme okay, not really important. Just merge it, or leave it be. I wanted to make sure methods work when adding no argument, which seems to be the case . . .

PR #37, opened 2018-10-27
WIP: Add shell_execute and properties list for given object

benlorenz:
For tab completion of given perl::Object, with a given prefix string which can be empty.
Note that this will not list methods like VISUAL, they will probably come via the parser and we need to merge them into the list for the output.
(If needed I might also look at other ways to get the list of possible methods)

I also added shell execute which was needed to add a custom rule file, this way we can define arbitrary polymake code (as polymake 'function') which can then be used via call_function as long as we can handle the return-type.

For the tab completion the return is Array<std::string>, i.e. this needs Mareks [enh/array_string](https://github.com/kalmarek/Polymake.jl/tree/enh/array_string) branch, then we can add an explicit to_array_string for the julia function, for now it is just an opaque PropertyValue.

### DIFF
--- a/deps/rules/julia.rules
+++ b/deps/rules/julia.rules
@@ -1,0 +1,6 @@
+
+function complete_property($$) {
+   my ($po,$prefix) = @_;
+   return new Array<String>(Core::Shell::Helper::try_property_completion($po->type,$prefix));
+}
+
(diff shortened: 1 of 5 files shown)


benlorenz:
I will do some more improvements soon (in a new pull request) as it doesn't really show the properties as I have just noticed but it should be fine for now.

PR #38, opened 2018-10-27
WIP: restructure propertyvaluehelper and add is_boolean

benlorenz:
Dont merge yet!

- Moved most of the functionality out of the PropertyValueHelper except for what we might add in the polymake core
- Added is_boolean and enabled to_bool in that case
- Switch demangling to typeinfo->name and __cxa_demangle instead of the polymake legible_typename function which might change (and as recommended by Ewgenij).
- Make the demangle step optional.

For the type-map we will at some point need both libc++ and libstdc++ variants since the internal names of all the std:: types differ between the c++ libraries (and we need at least std::string and std::pair).

### DIFF
--- a/deps/src/polymake.cpp
+++ b/deps/src/polymake.cpp
@@ -14,6 +14,8 @@
 
 #include "polymake_caller.h"
 
+#include <cxxabi.h>
+
 Polymake_Data data;
 
 JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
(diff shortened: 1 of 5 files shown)


sebasguts:
This looks good, thank you very much.

I would move the large lambda in `polymake.cpp` into a separate function in `polymake_functions.cpp`, if you don't mind, but that is just bikeshedding. Specially, I think most of the `polymake.method...` declarations from `polymake.cpp` should go there (some of them can be removed, too).

Is there a reason to not merge this, except for the different versions of the c++ libs? Maybe we could except those types on the c++ level and then return our own typename?

benlorenz:
I have moved the main function to polymake_functions and rebased onto the current master.
But I get some segfaults now which I need to investigate...

Regarding the types, we will probably need to do some regexping which I think we should do on the julia side and not in C++.
As a simple example some basic types after demangling the typeid->name:
```c++
    std::list<int>
    std::list<std::string>
    std::string
    std::pair<int,bool>
```
using libstdc++
```c++
std::__cxx11::list<int, std::allocator<int> >
std::__cxx11::list<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >
std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >
std::pair<int, bool>
```
and using libc++
```c++
std::__1::list<int, std::__1::allocator<int> >
std::__1::list<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >
std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >
std::__1::pair<int, bool>
```

The string could appear somewhere withing polymake types and the std::list could even be inside a polymake type and contain polymake types.

benlorenz:
Looking at this again, I think there are two options:
- either we try whether just removing "__cxx11::" and "__1::" suffices (it at least looks like that)
- generating our own names definitely need some regexp because we would need to find matching angle brackets and probably remove the traits and allocators

Since the first option should be really easy I will add this and we can think about the second later.

We might even auto-generate a small c++ test-program during build that when run outputs the mapping in whatever format needed for julia, i.e.
```C++
   int status = 1;
   std::cout << "std::list<int> -- "  << abi::__cxa_demangle(typeid(std::list<int>).name(),NULL,NULL,&status) << std::endl;
```

sebasguts:
I like the idea of having the small programm at build time. That way we have the correct types we need and can replace them without having to maintain a regex or assuming something about the libc++.

kalmarek:
Yet another idea is to do so at runtime (but ony once per julia session) by producing a `PropertyValue` of the known content and grabbing its `typeinfo_string`.

a julias `@generated` function could be used for this

sebasguts:
@benlorenz What is the current status of this?

benlorenz:
> @benlorenz What is the current status of this?

I was quite busy with trying to get BinaryBuilder working for polymake.
@kalmarek will try to write the type-map generator.

PR #40, opened 2018-10-29
catch missing "generated/includes.jl"

kalmarek:


saschatimme:
I am not really happy about this setup. Somehow this whole code generation should be done in a metaprogramming style and relying on precompilation to cache the generated functions.

saschatimme:
But obviously this fix is better than just throwing an error.

sebasguts:
@saschatimme we could definitely do this, I just do not see any benefit. What is a potential benefit?

PR #42, opened 2018-10-29
rework converting from PropertyValue

kalmarek:
we can have a discussion here about this;

The other options:
 * `WrappedTypes` is `Dict{Symbol,Type}` and we obtain `T` which we use to `convert(T, obj::PropertyValue)
 * some `@generated` function which automatically populates another `Dictionary{Symbol, Function}` at runtime
 * any other fancy meta-stuff which I have no idea about

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -39,58 +39,39 @@ function perlobj(name::String, input_data::Pair{Symbol}...; kwargsdata...)
     return obj
 end
 
-function typename_func(typename::String)
-    if typename == "int"
-        return to_int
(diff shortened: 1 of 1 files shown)


kalmarek:
well it had DONOTMERGE label :P

sebasguts:
Argh, sorry, I was too quick. You can revert it if you want to, and we discuss. But I am pretty happy about it for now.

kalmarek:
I'll open an issue then

sebasguts:
Fair enough, because the PR looked fine to me.

PR #47, opened 2018-11-03
WIP: add repl, access via $ key

benlorenz:
Probably needs some polishing as it was created for a pre-0.7 julia but seems to work after a few minor adaptions.
Also some tests would be nice.

I have used the jlcxx::make_julia_array for mixed arrays with bools, ints and strings (since tuples cause problems in CxxWrap.jl...), which is somewhat ugly but works.


sebasguts:
Stupid question, but is `$` really a good choice? Can't we set it to a way more uncommon combination, say `ctrl+alt+p`?

deps/src/polymake.cpp
```diff
@@ -147,13 +147,25 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
...
+
+  polymake.method("shell_complete",[](const std::string x)
+    {
+      auto res = data.main_polymake_session->shell_complete(x);
+      std::vector<std::string> props = std::get<2>(res);
+      jl_value_t** output = new jl_value_t*[props.size()+1];
+      output[0] = jl_box_int64(std::get<0>(res));
+      for (int i = 0; i < props.size(); ++i)
+         output[i+1] = jl_cstr_to_string(props[i].c_str());
+      return jlcxx::make_julia_array(output,props.size()+1);
```
sebasguts: I checked and made sure that this does not cause a memory leak.

benlorenz:
> Stupid question, but is `$` really a good choice? Can't we set it to a way more uncommon combination, say `ctrl+alt+p`?

I dont really care about the specific key, but I think a proper character is a lot better than some obscure control character:
 - This fits better to all the other repl-modes (; ] ?)
 - Control characters is always messy, it might be difficult to find on a mac or it might already be used by the terminal or readline library. (ctrl+p for example is used as an alternative for arrow up).

The dollar sign was just the canonical choice for polymake/perl.

src/Polymake.jl
```diff
@@ -73,6 +73,18 @@ include("functions.jl")
...
+
+module PolymakeREPLInit
+    using Polymake
+    using Polymake.PolymakeREPL
+    function __init__()
+        if isdefined(Base, :active_repl)
+            PolymakeREPL.RunPolymakeREPL()
+        end
+    end
+end
```
saschatimme: can you move this module also in the repl.jl file?

src/repl.jl
```diff
@@ -0,0 +1,102 @@
...
+
+
+function CreatePolymakeREPL(; prompt = "polymake > ", name = :pm, repl = Base.active_repl, main_mode = repl.interface.modes[1])
+   mirepl = isdefined(repl,:mi) ? repl.mi : repl
+   # Setup polymake panel
+   panel = LineEdit.Prompt(prompt;
+        # Copy colors from the prompt object
+        prompt_prefix=Base.text_colors[:yellow],
+        prompt_suffix=Base.text_colors[:white],
+        on_enter = REPL.return_callback)
```
saschatimme: I think this needs to check whether there is a `;` at the end.

saschatimme:
I simplified things a little bit (all these modules are not really necessary). Only thing I noticed is that you cannot enter multiline statements currently (Also see inline statement).

PR #52, opened 2018-11-07
Overload getindex for pm_perl_Object to make method call

saschatimme:
This is mostly a workaround until we get a working parser, but I think it's quite handy to get around any unimplemented features:

```julia
julia> c = cube(5);

julia> c[:DIM]() # this creates a closure and is immediately called
5

julia> c[:DIM]
#12 (generic function with 1 method)
```

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -41,7 +41,7 @@ end
 
 const WrappedTypes = Dict(
     Symbol("int") => to_int,
-    Symbol("double") => to_double, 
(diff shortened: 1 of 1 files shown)


kalmarek:
I understand that we workaround function vs method split for DIM in polymake
so what are other apparent uses apart from `DIM`?

saschatimme:
Pushed to the wrong remote, and now I cannot update this anymore...

PR #53, opened 2018-11-07
Overload getindex for pm_perl_Object to make method call

saschatimme:
This is mostly a workaround until we get a working parser, but I think it's quite handy to get around any unimplemented features:

```julia
julia> c = cube(5);

julia> c[:DIM]() # this creates a closure and is immediately called
5

julia> c[:DIM]
#12 (generic function with 1 method)
```

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -41,7 +41,7 @@ end
 
 const WrappedTypes = Dict(
     Symbol("int") => to_int,
-    Symbol("double") => to_double, 
+    Symbol("double") => to_double,
     Symbol("perl::Object") => to_perl_object,
     Symbol("pm::Integer") => to_pm_Integer,
     Symbol("pm::Rational") => to_pm_Rational,
@@ -55,7 +55,7 @@ const WrappedTypes = Dict(
     Symbol("pm::Array<long>") => to_array_int64,
     Symbol("pm::Array<std::basic_string<char,std::char_traits<char>,std::allocator<char>>>") => to_array_string,
     Symbol("pm::Array<pm::Set<int,pm::operations::cmp>>") => to_array_set_int32,
-    Symbol("pm::Array<pm::Matrix<pm::Integer>>") => to_array_matrix_Integer, 
+    Symbol("pm::Array<pm::Matrix<pm::Integer>>") => to_array_matrix_Integer,
     Symbol("undefined") => x -> nothing,
 )
 
@@ -71,7 +71,7 @@ function convert_from_property_value(obj::Polymake.pm_perl_PropertyValue)
     if haskey(WrappedTypes, T)
         f = WrappedTypes[T]
         return f(obj)
-    else    
+    else
         @warn("The return value contains $(typeinfo_string(obj)) which has not been wrapped yet")
         return obj
     end
@@ -82,7 +82,7 @@ function give(obj::Polymake.pm_perl_Object,prop::String)
         internal_give(obj, prop)
     catch ex
         throw(PolymakeError(ex.msg))
-    end 
+    end
     return convert_from_property_value(return_obj)
 end
 
@@ -96,3 +96,11 @@ function Base.show(io::IO, ::MIME"text/plain", obj::SmallObject)
     print(io, show_small_obj(obj))
 end
 Base.show(io::IO, obj::SmallObject) = show(io, MIME("text/plain"), obj)
+
+function Base.getindex(obj::pm_perl_Object, method::Symbol)
+    (args...) -> begin
+        convert_from_property_value(
+            call_method(string(method), obj, Array{Any,1}([args...]) )
+        )
+    end
+end


fingolfin:
Have you considered overloading `getproperty` instead, so that you can write `c.DIM()`?

kalmarek:
@fingolfin it's already there: `Base.getproperty(obj::pm_perl_Object, prop::Symbol) = give(obj, string(prop))`

sebasguts:
@fingolfin `getproperty` is already overloaded for properties. Polymake distinguishes between `methods`, `functions`, and `properties` on the C level. The problem is that some things that look like `properties` in the perl shell are in reality `methods`. That is what @saschatimme is trying to emulate. I am not sure this is the right thing to do here through.

saschatimme:
We can also just have a helper method which accomplishes the same thing. I think we just need some fallback for things not covered by our code generation stuff.

sebasguts:
@saschatimme The fallback is what you wrote inside the function. All possible methods will be covered once we have the code generation. Why am I against it:
* The distinctions between properties and methods is arbitrary already. It feels strange why on the polymake shell `VERTICES` and `DIM` feel the same, whereas in Julia, I need `.` for one and `[]` for another, and then there are even the functions. Unifying the way methods and functions are implemented seems like a bit less chaos
* Furthermore you are introducing some kind of member function syntax in Julia. This seems a bit against the language, doesn't it?

I suggest we stick with `polytope.DIM(c)`. We are not going to need a fallback really.

saschatimme:
I think `method(obj::pm_perl_Object, method_name::Symbol, args...)` would be the nicer alternative.
So `DIM` would then be `method(c, :DIM)`

sebasguts:
I find it way less suggestive than just `Polymake.method.DIM(c)` and `Polymake.func.minkowski_sum`. The `:DIM` is not an option, but something very important.
@fingolfin @benlorenz @micjoswig should maybe have a comment on which syntax to use, so we can have a vote.

fingolfin:
I certainly would prefer `blabla.DIM(c)`, also for tab completion purposes. Actually, what I'd really prefer (in terms of "make it Julia-like") is `import Polymake: dim; dim(c)`, but that's another step, I guess. 

That said, I am not currently using Polymake, so my voice should have little weight on this.

fingolfin:
As a non-polymake user, here is another simple minded question (please forgive my ignorance): How important is it that the user know about (resp. is exposed to) the difference between "funcs" and "methods"? Can there ever be funcs and methods that have the same name? Is this recommend / normal by Polymake standards?

Just thinking in terms of user friendliness /  discoverability, I imagine that in order to find things, I'd have to know by rote learning whether they are in `methods` or `func`, or else have to resort to trial-and-error. From that point of view, it'd be nicer if there was a single hierarchy. But of course that may not be feasible or even sensible for technical reasons.

sebasguts:
@fingolfin It will also be accessible via

```
import Polymake: polytope

polytope.DIM(c)
polytope.cube(4)
```

So there will be no distinction between methods and functions on this path. The `polytope` however is important, because Polymake organizes its functions in `applications`, and `polytope` is the one those functions and methods belong to. I think there cannot be methods and functions that have the same name in a single application, however, there can be same named functions in different applications.

sebasguts:
For the fallback solution proposed above the distinction is important. On the c++ level, methods are called via `c.method()` whereas functions are called via `function(c)` (makes sense from the name). So yes, the user needs to know this if he wants to use the fallback. I hope the fallback is not necessary anymore soonish.

benlorenz:
> I think there cannot be methods and functions that have the same name in a single application, however, there can be same named functions in different applications.

The application namespace is very important, there are several functions (and even big object types) with identical names in the different applications, for example:
```perl
polytope > print topaz::simplex(3)->type->full_name;
SimplicialComplex
polytope > print polytope::simplex(3)->type->full_name;
Polytope<Rational>
```
Multiple functions with the same name in a single application are also possible but they need to have different signatures for the overloading to work.

Having a function and a method of the same name is also possible, I don't know whether it is used for big objects at the moment but it is for small objects: For a matrix `cols($m)` will return an iterator over the columns but `$m->cols()` will return just the number of columns as an int.

benlorenz:
I would prefer if the methods could be used just like the properties, i.e. via `c.DIM` (maybe with `c.DIM()`). Otherwise this would be quite confusing for anyone who has previously used polymake. They are supposed to mimic the properties but need to be different mostly for technical reasons.

We could do this via a lookup within the give function, either by checking if this is a valid property (we already have a function that produces a list of all properties of a given object), or by generating a list of possible methods when reading the json.

sebasguts:
@benlorenz good to know, for this changes everything I guess.

I will try to do it via the `.` syntax.

What is your opinion on the `()`. I would guess there are methods that need `()`, and others not? Are they just optional?

micjoswig:
Currently, in poly make the user cannot feel the distinction between methods and functions (unless she looks at the documentation or the code).  I guess, on the current technical level this distinction was deemed convenient by the developers of the interface.  Let us not bother about this for now. My idea would have such issues resolved by one higher level of abstraction. 

— Michael


Michael Joswig
Institut für Mathematik, MA 6-2
TU Berlin

benlorenz:
> What is your opinion on the `()`. I would guess there are methods that need `()`, and others not? Are they just optional?

When there are no additional arguments given the parentheses are optional and the method is called with no extra arguments, `$c->DIM;` and `$c->DIM();` does exactly the same.
But there are also methods like `$c->N_LATTICE_POINTS_IN_DILATION(3);` which need arguments and some many like `$c->VISUAL(FacetColor=>"blue");` where we can pass extra arguments but `$c->VISUAL;` also works.

sebasguts:
@micjoswig @benlorenz I do not really understand, because your comments seem to contradict each other. @benlorenz said methods and functions with the same name can have different semantics. @micjoswig says that polymake users cannot feel the difference. What is true here?

On the Julia level, how can we distinguish a hypothetical function `DIM` from a hypothetical method `DIM`, both called on a polytope. Currently, this would cause the generation of functions by our parser to crash.

micjoswig:
Sorry. My bad. Of course, Benjamin is right. I was referring to methods and 
user_methods.

Am 7. November 2018 11:35:54 schrieb Sebastian Gutsche 
<notifications@github.com>:

sebasguts:
Okay, so what we have to deal with are

    Properties: accessed in polymake via ->FOO, in Julia via .FOO, on the C++ level via give
    Methods: accessed in polymake via ->FOO to mock properties or via ->foo(), accessed in julia via application.foo() and on the C++ level via x.prepare("foo").
    Functions: accessed in polymake via foo(x), in Julia via application.foo(x), and on the C++ level via "prepare("foo")`

So far the current (not quite nice situation).

I would proceed to do the following

    Make functions callable exactly like they are now.
    As proposed by @benlorenz, make a single method module, so methods are generally accessed via method.foo(x)
    Make the properties accessible via x[:FOO]
    For single argument, allow the x.FOO syntax via a trick: Store those method names in a list, and if someone types in x.FOO, look up if FOO is in that list, and if so, call the method directly (no brackets), or dispatch to x[:FOO].

This way, a user that knows what he is doing can avoid the lookup, whereas the polymake syntax is still quite similar to polymake for people just coming over.

What do you think?

saschatimme:
> I would proceed to do the following
> 
> * Make functions callable exactly like they are now.

agreed

> * Make the properties accessible via `x[:FOO]`
> * For single argument, allow the `x.FOO` syntax via a trick: Store those method names in a list, and if someone types in `x.FOO`, look up if `FOO` is in that list, and if so, call the method directly (no brackets), or dispatch to `x[:FOO]`.

Can't these two be overloaded by the same syntax `x.FOO` syntax? Basically if we know it's not a method, just use property access

> * As proposed by @benlorenz, make a single `method` module, so methods are generally accessed via `method.foo(x)`

I am not a huge fan of this and personally think that the `method(x, :foo, args...)` syntax is nicer / more Julian. But that's obviously bike shedding. But to make clear, this is intended as an **fallback** and all methods *should* be covered by code generation.
> 
> This way, a user that knows what he is doing can avoid the lookup, whereas the polymake syntax is still quite similar to polymake for people just coming over.
> 
> What do you think?

sebasguts:
> Can't these two be overloaded by the same syntax x.FOO syntax? Basically if we know it's not a method, just use property access

Yes, that's what I said. I just wanted to have ways to call properties and methods separately, too. The `.` accessor will then just dispatch on properties and methods

I think it was not really clear what I meant. "For single argument" means "For single argument methods"

PR #54, opened 2018-11-07
Update Travis / Codecov / Coveralls integration; drop AppVeyor; require Julia >= 1.0

fingolfin:
This is not yet actually working, because key parts are missing (setting
up a suitable Polymake ...) but at least it is somewhat closer to being
useful than it was before.


.appveyor.yml
```diff
@@ -0,0 +1,43 @@
+environment:
+  matrix:
+  - julia_version: 0.7
+  - julia_version: 1
+  - julia_version: nightly
+
+platform:
+  - x86 # 32-bit
```
saschatimme: We won't work on 32-bit

.appveyor.yml
```diff
@@ -0,0 +1,43 @@
+environment:
+  matrix:
+  - julia_version: 0.7
```
saschatimme: I think 1.0 only is fine by now

fingolfin: Well, I matched what you declare in `REQUIRE`... I think the two should match, else you make a promise that is not verified. So, which is it -- drop 0.7 here and in `REQUIRE`, or keep it in both?

.appveyor.yml
fingolfin: OK, will change.

But is Windows supported at all? If not, perhaps this should be removed completely...

saschatimme: No not really, just WSL so in the end only UNIX systems... So just drop this :)

.appveyor.yml
saschatimme: Good point, we should bump the REQUIRE.

.appveyor.yml
sebasguts: @fingolfin Polymake does not work on Windows at all, so no need to support windows.

fingolfin: Excellent. I was surprised to see the AppVeyor config in here, happy to remove it instead of updating it :)

.appveyor.yml
fingolfin: The drawback of that is that if any other parts of OSCAR will require 0.7 for some time still, they won't be usable together with Polymake.jl... Is there a downside to keeping 0.7 support in for now?

Perhaps you are more confident than me that all parts of OSCAR will "soon" support Julia 1.0?

saschatimme: If there is a need to still support 0.7, I think we can also do this. But things like `AbstractAlgebra.jl` doesn't work with 0.7 (there is a compiler bug in 0.7) so I don't see how OSCAR will run under 0.7 :)

fingolfin: Well, I am forced to run under Julia master branch, because 1.0 is missing crucial infra structure for GAPJulia, so I guess you are right. I was mostly thinking about Singular.jl, but I guess we'll simply have to wait for them to support 1.0 / 1.1.

Will drop 0.7 from here and REQUIRE

fingolfin:
Updated, and as you can see, I enabled Travis for this repo (actually, for this whole organization). Of course right now it is always failing, because there is no polymake installed. And adding that to Travis may require some care if it meant to be efficient. Perhaps there is a pre-built Polymake binary somewhere that can be used? Otherwise, it might be possible to build Polymake if one is careful to use Travis' caching feature and an incremental build (to reduce the time for each Travis build to a reasonable amount). But I am not sure how reliable incremental builds of Polymake are? (in GAP, I worked hard to make them super reliable, but people still don't trust them and do `make clean && make all` all the time, because they were burned so often in the past; and Julia is a disaster in this regard, too, at least when any of its builtin deps change).

saschatimme:
@benlorenz is working on a BinaryBuilder version of Polymake, which we then could use. So far this is not ready for prime time (but we are getting close...)

saschatimme:
@fingolfin can we enable on a per repo basis and disable it for Polymake.jl for now?

fingolfin:
Sure, no problem, I have disable Travis builds for Polymake.jl again for now (it was simply necessary to temporarily enable them to make sure the changes in this PR work -- which I now know).

PR #55, opened 2018-11-07
Added helper to call methods/functions

sebasguts:
This is an alternative to #53 

Calling looks like
```
julia> c = Polymake.func.cube(3);

julia> Polymake.method.DIM(c)
3
```

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -96,3 +96,34 @@ function Base.show(io::IO, ::MIME"text/plain", obj::SmallObject)
     print(io, show_small_obj(obj))
 end
 Base.show(io::IO, obj::SmallObject) = show(io, MIME("text/plain"), obj)
+
+
+
+struct PolymakeFunctionType
+end
+
+const func = PolymakeFunctionType()
+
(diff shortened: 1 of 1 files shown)


benlorenz:
When we make the functions available via this syntax I would prefer to have the functions directly under the application namespace and the methods could be grouped under the corresponding big object types:
```julia
julia> c = Polymake.polytope.cube(3);

julia> Polymake.polytope.Polytope.DIM(c)
3
```

kalmarek:
`Polymake.polytope.Polytope.DIM(c)` is pretty horrible :-)
Not to mention that a user would expect the dimension function of a polytope to live in `Polymake.polytope` space

sebasguts:
@benlorenz actually, this idea is not too bad. It looks horrible at first, but @kalmarek remember, you can still `using polytope` and then write `Polytope.DIM`

benlorenz:
> `Polymake.polytope.Polytope.DIM(c)` is pretty horrible :-)
> Not to mention that a user would expect the dimension function of a polytope to live in `Polymake.polytope` space

We could also group them together under `.method` instead of separating them by object types since polymake will do the overload resolution anyway.

PR #56, opened 2018-11-07
Add `call_method_void`

saschatimme:
Polymake is quite opinionated when to show a visualization. In particular the method has to be called in a void context. We therefore need a new `call_method_void` which throws away the return value...

With this
```julia
julia> c = cube(3);

julia> Polymake.call_method_void("VISUAL", c, [])
polymake: used package jreality
  jReality is a visualization and sonification library focused on mathematical and scientific content.
  http://www3.math.tu-berlin.de/jreality
```
works :)

@benlorenz figured this out :)

How we expose this to the user is another question and there is already enough bike shedding today, so I will punt on this for now...

### DIFF
--- a/deps/src/polymake_caller.cpp
+++ b/deps/src/polymake_caller.cpp
@@ -97,7 +97,6 @@ void polymake_call_method_void(std::string function_name, pm::perl::Object objec
         polymake_call_function_feed_argument(function, arguments[i]);
     }
     function();
-    return;
 }
 
 
--- a/deps/src/polymake_caller.cpp
+++ b/deps/src/polymake_caller.cpp
@@ -89,8 +89,21 @@ pm::perl::PropertyValue polymake_call_method(std::string function_name, pm::perl
     return function();
 }
 
+void polymake_call_method_void(std::string function_name, pm::perl::Object object, jlcxx::ArrayRef<jl_value_t*> arguments)
+{
+    size_t argument_list = arguments.size();
+    auto function = object.prepare_call_method(function_name);
+    for(size_t i = 0;i<argument_list;i++){
+        polymake_call_function_feed_argument(function, arguments[i]);
+    }
+    function();
+    return;
+}
+
+
 void polymake_module_add_caller(jlcxx::Module& polymake){
     polymake.method("call_function",&polymake_call_function);
     polymake.method("call_method",&polymake_call_method);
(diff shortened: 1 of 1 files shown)


sebasguts:
This looks good, thanks a lot. I think exposing this to the user is hard, as the proposed JSON will not give us information about the output.

@benlorenz Do you have an idea?

sebasguts:
Okay, so what we have to deal with are
* Properties: accessed in polymake via `->FOO`, in Julia via `.FOO`, on the C++ level via `give`
* Methods: accessed in polymake via `->FOO` to mock properties or via `->foo()`, accessed in julia via `application.foo()` and on the C++ level via `x.prepare("foo")`.
* Functions: accessed in polymake via `foo(x)`, in Julia via `application.foo(x)`, and on the C++ level via "prepare("foo")`

So far the current (not quite nice situation).

I would proceed to do the following

* Make properties and functions callable exactly like they are now.
* As proposed by @benlorenz, make a single `method` module, so methods are generally accessed via `method.foo(x)`
* Make the properties accessible via `x[:FOO]`
* For single argument, allow the `x.FOO` syntax via a trick: Store those method names in a list, and if someone types in `x.FOO`, look up if `FOO` is in that list, and if so, call the method directly (no brackets), or dispatch to `x[:FOO]`.

This way, a user that knows what he is doing can avoid the lookup, whereas the polymake syntax is still quite similar to polymake for people just coming over.

What do you think?

deps/src/polymake_caller.cpp
```diff
@@ -89,8 +89,21 @@ pm::perl::PropertyValue polymake_call_method(std::string function_name, pm::perl
...
 
+void polymake_call_method_void(std::string function_name, pm::perl::Object object, jlcxx::ArrayRef<jl_value_t*> arguments)
+{
+    size_t argument_list = arguments.size();
+    auto function = object.prepare_call_method(function_name);
+    for(size_t i = 0;i<argument_list;i++){
+        polymake_call_function_feed_argument(function, arguments[i]);
+    }
+    function();
+    return;
```
sebasguts:
```suggestion
```
The return here is unnecessary

saschatimme: Here you see that I have no idea about C++^^. I need a new approval now :)

PR #57, opened 2018-11-12
added functions to export list

micjoswig:


sebasguts:
@micjoswig Thanks, I missed those a long time ago.

Please note that those functions were implemented by me to restore the functionality you had demonstrated in your example notebook, when we switched from Cxx to CxxWrap. They will (hopefully soonish) disappear to the benefit of their `polytope....` counterparts, which will be automatically generated.

PR #58, opened 2018-11-13
add to_string for PropertyValue, used as fallback for print

benlorenz:
it turns out that any PropertyValue (probably except for BigObject and related types) can be converted to std::string, which we can use as fallback for printing unwrapped types:
```
julia> cube(3).FACETS
┌ Warning: The return value contains pm::SparseMatrix<pm::Rational, pm::NonSymmetric> which has not been wrapped yet
└ @ Polymake ~/software/polymake/julia/Polymake.jl/src/functions.jl:75
1 1 0 0
1 -1 0 0
1 0 1 0
1 0 -1 0
1 0 0 1
1 0 0 -1
```

we should probably add some tests

### DIFF
--- a/deps/src/polymake.cpp
+++ b/deps/src/polymake.cpp
@@ -73,6 +73,9 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
   polymake.method("to_matrix_Rational", [](pm::perl::PropertyValue pv){
     return to_SmallObject<pm::Matrix<pm::Rational>>(pv);
   });
+  polymake.method("to_string", [](pm::perl::PropertyValue pv){
+    return to_SmallObject<std::string>(pv);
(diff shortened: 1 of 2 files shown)


kalmarek:
ok! maybe we should rename `to_SmallObject<T>` to `convert_to<T>` then?

sebasguts:
Looks good to me, thank you :)

benlorenz:
> ok! maybe we should rename `to_SmallObject<T>` to `convert_to<T>` then?

I think convert is also somewhat misleading because it can do a conversion but in many cases it just extracts the pointer and does a reinterpret_cast without any conversion.

The internal methods in polymake are called `retrieve`.

sebasguts:
I would be fine with retrieve.

PR #59, opened 2018-11-14
Remove old call_func_*args functions and add new convenience call_func

saschatimme:
This came up in a discussion with Michael. We currently have some old `call_func_0args`, `call_func_1args` and `call_func_2args` functions lying around. I removed them and replaced them with a new varargs `call_func` method (mostly as an intermediate step until we covered everything by our code generation).

### DIFF
--- a/deps/src/polymake_functions.cpp
+++ b/deps/src/polymake_functions.cpp
@@ -16,18 +16,6 @@ void initialize_polymake(){
     }
 }
 
-polymake::perl::Object call_func_0args(std::string func) {
-    return polymake::call_function(func);
-}
-
-polymake::perl::Object call_func_1args(std::string func, int arg1) {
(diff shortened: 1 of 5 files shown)


kalmarek:
looks good to me;

maybe we could have a test with `call_func(::String, ...)` or two?

saschatimme:
I added some tests. I didn't remove them, since Michael currently uses them in a demo and I also don't know how the json should look. Can you help me there?

sebasguts:
@saschatimme Sure, will do that tonight.
@micjoswig You will need to rearrange a tiny bit in your demo, I hope that is okay.

sebasguts:
@saschatimme On second thought, this can also be done in a separate PR then, no need to keep this unmerged.

saschatimme:
Great, thanks!

PR #60, opened 2018-11-14
Remove old call_func_*args methods on the cpp side

saschatimme:
I messed up in #59 and forgot to remove all appearances on the cpp side.


sebasguts:
:D I see I was late with #61

PR #62, opened 2018-11-14
Temp defs

sebasguts:
This PR contains #61

PR #63, opened 2018-11-16
remove entire export list; adapt functions to Julia-ish style

micjoswig:
add new function pseudopower

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -1,22 +1,14 @@
-export cube, cross, perlobj, rand_sphere, upper_bound_theorem, call_func
-
 import Base: convert, show
 
-function cube(dim)
-    return call_func(:cube, dim)
-end
+# application polytope
+cube(dim::Int) = call_func(:cube, dim)
+cross(dim::Int) = call_func(:cross, dim)
+rand_sphere(n::Int, d::Int) = call_func(:rand_sphere, n, d)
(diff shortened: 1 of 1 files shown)


src/functions.jl
```diff
@@ -1,22 +1,14 @@
...
-end
-
-function cross(dim)
-    return call_func(:cross, dim)
-end
+# application polytope
+cube(dim::Int) = call_func(:cube, dim)
+cross(dim::Int) = call_func(:cross, dim)
+rand_sphere(n::Int, d::Int) = call_func(:rand_sphere, n, d)
+upper_bound_theorem(n::Int, d::Int) = call_func(:upper_bound_theorem, n, d)
```
sebasguts: Those functions are obsolete with #62, I would suggest merging that one first

src/functions.jl
```diff
@@ -1,22 +1,14 @@
...
 
-function rand_sphere(n,d)
-    return call_func(:rand_sphere, n, d)
-end
-
-function upper_bound_theorem(n,d)
-    return call_func(:upper_bound_theorem, n, d)
-end
+# application common
+pseudopower(a::pm_Integer, b) = call_func(:pseudopower, a, b)
```
sebasguts: Those could be added better using JSON defs.

PR #68, opened 2018-11-25
WIP: Moved method definitions into separate module

sebasguts:
* Methods are no longer in the application module, but may be accessed via `methods.foo`
* Single-argument methods are now callable via `getproperty`
* Properties are accessible via `getindex`

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -79,6 +79,8 @@ function give(obj::Polymake.pm_perl_Object,prop::String)
     return convert_from_property_value(return_obj)
 end
 
+Base.getindex(obj::pm_perl_Object, prop::Symbol) = give(obj,string(prop))
+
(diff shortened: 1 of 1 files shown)


deps/parser/parser.jl
```diff
@@ -163,54 +162,96 @@ function parse_definition(method_dict::Dict, app_name::String)
...
         end
     end
-    return return_string
+    return function_string,method_string
 end
 
-function parse_app_definitions(filename::String,outputfileposition::String,include_file::String)
+module_string_includes = """
+
+import ..pm_Integer, ..pm_Rational, ..pm_Matrix, ..pm_Vector, ..pm_Set,
```
saschatimme: Can you explain why we need the imports here? Also I think this is not a great idea, since we need to keep all these exports and imports in sync and this will become a mess...

saschatimme:
The conclusion was now that we don’t do this, right?

PR #70, opened 2018-11-26
Add more Array wrappers

saschatimme:
I needed to wrap some more `pm:Array` instances.


kalmarek:
How was the experience of adding new types?
did You have a look at my guide, or was is evident enough?

saschatimme:
All these type tables are a mess. The rest I could do from my memory.

PR #71, opened 2018-11-26
Add temporary call_method

saschatimme:
This is a workaround until we figure this stuff out later this week.


saschatimme:
No, but I was also not able to find the others ones...

PR #72, opened 2018-11-26
Unify type translation lists

sebasguts:


kalmarek:
I like that, but it makes my guide obsolete ;-)

deps/parser/type_setup.jl
```diff
@@ -0,0 +1,53 @@
+type_tuples = Any[
```
saschatimme: No need to put an `Any` here

sebasguts:
I adressed all your comments
@kalmarek not obsolete, just shorter

PR #74, opened 2018-11-27
Applied clang-format to all cxx-files

sebasguts:


saschatimme:
Olivia currently cleans up the `deps/src/polymake.cpp` (moving stuff into their own files). I think we should wait until that is merged.

sebasguts:
Indeed, we should wait for that.

sebasguts:
@saschatimme I rebased that one. Can you maybe merge it?

PR #76, opened 2018-11-27
Binarybuilder

benlorenz:
first try at getting binaryprovider working for polymake, the build is based on the current snapshot
I have forked most of the dependencies because they were built too long ago, i.e. too old BinaryBuilder version which doesnt support CompilerABI stuff...

Cross compiling is not possible at the moment (for perl, boost and polymake), hence only x86 linux is available.

I also tried to enable travis.


saschatimme:
The travis-ci build comes with 
```
│ -- The C compiler identification is GNU 5.4.0
│ -- The CXX compiler identification is GNU 5.4.0
```
so, we probably need to install GCC I assume?

sebasguts:
Very nice!

PR #79, opened 2018-11-28
Add call_method helper function and improve call_function helper 

saschatimme:


deps/src/polymake_caller.cpp
```diff
@@ -24,6 +22,15 @@ void set_julia_type(std::string name, void* type_address)
     memcpy(address, &type_address, sizeof(jl_value_t*));
 }
 
+void* get_ptr_from_cxxwrap_obj(jl_value_t* obj)
```
sebasguts: This needs to be removed again.

PR #82, opened 2018-11-30
Visualization in IJulia

sebasguts:


sebasguts:
This should be finished now.

PR #83, opened 2018-11-30
WIP: Type translation function

sebasguts:
@saschatimme Can you check if this new function compiles and works on Mac please?


saschatimme:
Also works on my mac

benlorenz:
On 12/2/18 1:22 PM, Sebastian Gutsche wrote:
> @benlorenz I guess this is done for now, but needs #38 to work, so we get the real types, am I right?

Yes that is correct, the travis test probably failed because there is
::__cxx11 in all the generated std::strings that doesn't appear in the
legible_typename.

> So, if you agree, I would resourrect the necessary parts from #38 

Thanks!

sebasguts:
@benlorenz 
Made the merge and test pass now, lets hope they also pass on travis.

There is one minor thing. `typeinfo_string` returns `perl::Object` instead of `pm::perl::Object`. I currently left this as corner case. Any idea how we can solve that?

sebasguts:
@saschatimme This PR is now ready

deps/src/polymake_functions.cpp
```diff
@@ -19,9 +22,63 @@ void initialize_polymake()
...
+
+        // with typeinfo ptr (nullptr for Objects)
+        case PropertyValueHelper::number_is_object:
+            // some non-primitive Scalar type with typeinfo (e.g. Rational)
+        case PropertyValueHelper::not_a_number:
+            // a c++ type with typeinfo or a perl Object
+            {
+                const std::type_info* ti = ph.get_canned_typeinfo();
+                if (ti == nullptr) {
+                    return "perl::Object";
```
benlorenz: The string here does not contain the ```pm::```, adding it here should avoid the special case in the ```WrappedTypes```.

Also, when there is an object of type ```pm::perl::Object``` inside a ```PropertyValue``` there is no ```type_info``` struct, and we currently just guess this to be a ```pm::perl::Object```. (It might also be: pm::Array<pm::perl::Object>, pm::perl::ObjectType or some plain perl type, and maybe more...)
I'm not sure if there is a way to determine these types at runtime, we might need to use/trust the json for those.

src/functions.jl
```diff
@@ -26,41 +26,33 @@ end
...
-    Symbol("pm::Array<long>") => to_array_int64,
-    Symbol("pm::Array<pm::Integer>") => to_array_Integer,
-    Symbol("pm::Array<pm::Array<int>") => to_array_array_int32,
-    Symbol("pm::Array<pm::Array<long>") => to_array_array_int64,
-    Symbol("pm::Array<pm::Array<pm::Integer>") => to_array_array_Integer,
-    Symbol("pm::Array<std::basic_string<char,std::char_traits<char>,std::allocator<char>>>") => to_array_string,
-    Symbol("pm::Array<pm::Set<int,pm::operations::cmp>>") => to_array_set_int32,
-    Symbol("pm::Array<pm::Matrix<pm::Integer>>") => to_array_matrix_Integer,
     Symbol("undefined") => x -> nothing,
+    Symbol("perl::Object") => to_perl_object,
```
benlorenz: This special case can be removed, see above.

sebasguts:
@benlorenz I adressed your comments

sebasguts:
Thank you for the approval

PR #84, opened 2018-12-11
BigObjectArrays, types and more

benlorenz:
Several improvements and additions (@lkastner ):

- determine perl based types (without type_info) via perl helper function, this allows to check whether a PropertyValue is a big object, a big object array, or something else
- add arrays of big objects, which needed a special case since they dont have the same methods as other arrays. this should help for atint and polydb
- add ObjectType (this could be used for the perlobj calls as an alternative to the type string)
- add a few take methods and an add method for multiple subobjects
- minor fix in vector construction as suggested by @saschatimme 

no tests yet...

### DIFF
--- a/deps/src/polymake_arrays.cpp
+++ b/deps/src/polymake_arrays.cpp
@@ -83,32 +83,32 @@ void polymake_module_add_array(jlcxx::Module& polymake)
             });
         });
 
-    polymake.method("to_array_int32", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_int32", [](const pm::perl::PropertyValue& pv) {
         return to_SmallObject<pm::Array<int32_t>>(pv);
     });
-    polymake.method("to_array_int64", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_int64", [](const pm::perl::PropertyValue& pv) {
         return to_SmallObject<pm::Array<long>>(pv);
     });
-    polymake.method("to_array_Integer", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_Integer", [](const pm::perl::PropertyValue& pv) {
         return to_SmallObject<pm::Array<pm::Integer>>(pv);
     });
-    polymake.method("to_array_string", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_string", [](const pm::perl::PropertyValue& pv) {
         return to_SmallObject<pm::Array<std::string>>(pv);
     });
-    polymake.method("to_array_array_int32", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_array_int32", [](const pm::perl::PropertyValue& pv) {
         return to_SmallObject<pm::Array<pm::Array<int32_t>>>(pv);
     });
-    polymake.method("to_array_array_int64", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_array_int64", [](const pm::perl::PropertyValue& pv) {
         return to_SmallObject<pm::Array<pm::Array<long>>>(pv);
     });
-    polymake.method("to_array_array_Integer", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_array_Integer", [](const pm::perl::PropertyValue& pv) {
         return to_SmallObject<pm::Array<pm::Array<pm::Integer>>>(pv);
     });
-    polymake.method("to_array_set_int32", [](pm::perl::PropertyValue pv) {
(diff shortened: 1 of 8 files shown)


deps/src/polymake_arrays.cpp
```diff
@@ -57,34 +57,67 @@ void polymake_module_add_array(jlcxx::Module& polymake)
...
+                return A;
+            });
+            wrapped.method("resize!", [](WrappedT& A, int32_t newsz) {
+                A.resize(static_cast<long>(newsz));
+                return A;
+            });
         });
 
-    polymake.method("to_array_int32", [](pm::perl::PropertyValue pv) {
+    polymake.method("to_array_int32", [](const pm::perl::PropertyValue& pv) {
```
sebasguts: Good spot. Avoids lots of copies now :)

deps/src/polymake_perl_objects.cpp
```diff
@@ -82,6 +89,11 @@ void polymake_module_add_perl_object(jlcxx::Module& polymake)
     polymake.method("take",
                     [](pm::perl::Object p, const std::string& s,
                        const pm::perl::PropertyValue& v) { p.take(s) << v; });
+    polymake.method("take",
+                    [](pm::perl::Object p, const std::string& s,
+                       const pm::perl::Object& v) { p.take(s) << v; });
+    polymake.method("add", [](pm::perl::Object p, const std::string& s,
+                              const pm::perl::Object& v) { p.add(s, v); });
```
sebasguts: Is this the add method for `pm::Array`?

src/functions.jl
```diff
@@ -125,4 +125,7 @@ end
 function Base.show(io::IO, ::MIME"text/plain", pv::pm_perl_PropertyValue)
     print(io, to_string(pv))
 end
+function Base.show(io::IO, ::MIME"text/plain", a::pm_Array{pm_perl_Object})
+    print(io, "pm_Array{pm_perl_Object} of size ",length(a))
```
sebasguts: Should we in best Julia style recursively print the elements?

deps/src/polymake_arrays.cpp
```diff
@@ -57,34 +57,67 @@ void polymake_module_add_array(jlcxx::Module& polymake)
             wrapped.method("take",
                            [](pm::perl::Object p, const std::string& s,
                               WrappedT& A) { p.take(s) << A; });
+        })
+        .apply<pm::Array<pm::perl::Object>>([](auto wrapped) {
```
sebasguts: Nice, didn't know chaining the `apply`s works

benlorenz:
On 12/11/18 11:58 AM, Sebastian Gutsche wrote:
>> @@ -82,6 +89,11 @@ void polymake_module_add_perl_object(jlcxx::Module& polymake)
>      polymake.method("take",
>                      [](pm::perl::Object p, const std::string& s,
>                         const pm::perl::PropertyValue& v) { p.take(s) << v; });
> +    polymake.method("take",
> +                    [](pm::perl::Object p, const std::string& s,
> +                       const pm::perl::Object& v) { p.take(s) << v; });
> +    polymake.method("add", [](pm::perl::Object p, const std::string& s,
> +                              const pm::perl::Object& v) { p.add(s, v); });
> 
> Is this the add method for `pm::Array`?

No, this is for adding objects as multiple subobjects, for example
adding an LP to a polytope.

The Array can be assigned via setindex: arr[i]

>> @@ -125,4 +125,7 @@ end
>  function Base.show(io::IO, ::MIME"text/plain", pv::pm_perl_PropertyValue)
>      print(io, to_string(pv))
>  end
> +function Base.show(io::IO, ::MIME"text/plain", a::pm_Array{pm_perl_Object})
> +    print(io, "pm_Array{pm_perl_Object} of size ",length(a))
> 
> Should we in best Julia style recursively print the elements?

This can get very long with our rather verbose property-listing,
especially for polydb which is one of the main use cases.
We might print a list of names if the objects have those.

Also the objects might be invalid if the array is not yet filled
completely and printing those will cause a segfault.

I will try to come up with something nicer soon.

>> @@ -57,34 +57,67 @@ void polymake_module_add_array(jlcxx::Module& polymake)
>              wrapped.method("take",
>                             [](pm::perl::Object p, const std::string& s,
>                                WrappedT& A) { p.take(s) << A; });
> +        })
> +        .apply<pm::Array<pm::perl::Object>>([](auto wrapped) {
> 
> Nice, didn't know chaining the `apply`s works

@saschatimme suggested to try this.

sebasguts:
@benlorenz okay, looks very good then. Thank you :)

PR #85, opened 2019-01-07
json generation

benlorenz:
first version of json generator script  (@lkastner, @sebasguts )

tests currently fail:
 - the cube function as documented in expects a template parameter, I'm not sure how to detect whether this is a mandatory or optional template argument
 - pseudopower is a function not a user_function and thus does not appear in the json


sebasguts:
@benlorenz Looks good. Will compile a new polymake to test it. The following questions:

Regarding 1: Until now, we did not specify a type parameter for `cube`. I would guess `Rational` is just a default one? The JSON described [here](https://trac.polymake.org/ticket/1094) just has `type_params`, which is specified there as "mandatory type parameters". So we might need two numbers, i.e., `mandatory_type_params` and `type_params`?

Regarding 2: What is the difference between a function and a user function? Do we need to create separate JSON files for functions, or can they not be exported, as they lie deeper in the system?

benlorenz:
> @benlorenz Looks good. Will compile a new polymake to test it. The following questions:
> 
> Regarding 1: Until now, we did not specify a type parameter for `cube`. I would guess `Rational` is just a default one? The JSON described [here](https://trac.polymake.org/ticket/1094) just has `type_params`, which is specified there as "mandatory type parameters". So we might need two numbers, i.e., `mandatory_type_params` and `type_params`?

This would help, but the information doesn't exist in the polymake help format. I will have a look if we can get this from somewhere else or if we will need improve the polymake documentation.
This can be observed in the polymake shell as well: typing `$p=cube` and pressing `<TAB>` will append a `<` because polymake expects a template parameter here even though it is not needed.

> Regarding 2: What is the difference between a function and a user function? Do we need to create separate JSON files for functions, or can they not be exported, as they lie deeper in the system?

Technically they are the same except for the doc-string, and because of this `functions` are not shown in tab-completion, online-help or documentation; they are used internally and usually not supposed to be called by users. (but they can be used if you know the name)
The way the json generator works at the moment there is no way to list the non-user functions. 
So there are a few options:
 - `pseudopower` could be made a `user_function` within polymake
 - `Polymake.jl` could use `call_function` instead
 - we could add a manually generated json file for extra functions where we want to have a generated julia function


It seems running polymake for the json generator doesn't seem to work with the binary-builder polymake (on travis), I will have a look at that within the next days.

sebasguts:
> This would help, but the information doesn't exist in the polymake help format. I will have a look if we can get this from somewhere else or if we will need improve the polymake documentation.

Okay. I guess it is an interesting information for the documentation itself. I guess it is always just a default value that is inserted, and no significant behavioural changes happen when a different number of type params is added? And maybe there is always a default? That way we could actually assume that we can always make it optional?

> they are used internally and usually not supposed to be called by users. (but they can be used if you know the name)

Then I would say this is an expert thingy, and those functions should either be used using `call_func` and `call_method`, or we add a manual extracted sublist to an extra json. Maybe it would even be useful to look what people want and make them user functions at some point.

Another problem occured to me: Currently, `cube` is called like this:
```
internal_call_function( "cube<$param1>", Any[ arg1, ] )
```
Afaik, this did not work (see email from Eugen, October 1, 2018, JSON für Julia). Do you know if this `mlist` construction there ever got implemented in polymake? Or how this was solved?

sebasguts:
> We might do some guessing by checking whether the type appears in the types for the parameters.

For the cube, the type param only appears in the optional arguments. So should we check them, too? May it happen that there is a function for which an optional argument may set the type param, but when only required parameters are used, it needs to be set?

> I think having an extra json file for such functions is a good solution.

All right, so we will maintain a manually written JSON. The question is how we should name this module. Currently, functions appear in a module which is named after their application, i.e., `cube` is called via `polytope.cube`. Should we just name them `internal` or `auxiliary`? I would probably vote for the latter.

> The `type_parameters("Max")` version using a string which Ewgenij suggested doesn't exist yet.

But we need this for the type params to work correctly anyway, right?

sebasguts:
I removed the type parameters for now. They can be readded once we have a way to actually pass them to functions.

There now is a folder `JSON` in which custom JSON's can be added. They are parsed along with the generated JSON

sebasguts:
@benlorenz Could you have a look at the changes?

PR #88, opened 2019-01-18
Some bugfixes

sebasguts:
Some bugfixes in function.jl and convert.jl

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -98,12 +98,12 @@ function convert_matrix_rational(pmmatrix::pm_perl_PropertyValue)
 end
 
 function convert_matrix_rational(matrix::pm_Matrix{pm_Rational})
-    rows = rows(matrix)
+    nr_rows = rows(matrix)
     columns = cols(matrix)
-    result = Array{Rational{BigInt},2}(rows,columns)
-    for i = 1:rows
+    result = Array{Rational{BigInt},2}(undef,nr_rows,columns)
+    for i = 1:nr_rows
         for j = 1:columns
-            current_entry = matrix(i-1,j-1)
+            current_entry = matrix[i,j]
             result[i,j] = convert(Rational{BigInt},current_entry)
         end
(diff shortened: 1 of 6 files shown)


kalmarek:
we should have tests for indexing/converts into Matrices/Vectors
If I remember polymake is row-major, so converts are going to be inefficient.
 
@saschatimme how about linearindexing (so that we could `for i in eachindex`) them? worth the effort?

saschatimme:
> If I remember polymake is row-major, so converts are going to be inefficient.
C/C++ is row major, so polymake as well :)

> How about linearindexing (so that we could for i in eachindex) them? worth the effort?
What effort? 
```julia
eachindex(A::AbstractArray) = (@_inline_meta(); eachindex(IndexStyle(A), A))
eachindex(::IndexLinear, A::AbstractArray) = (@_inline_meta; OneTo(length(A)))
eachindex(::IndexLinear, A::AbstractVector) = (@_inline_meta; axes1(A))
```
(btw you can find source code super easy with the `@less` macro, e.g. `A=rand(2,3);@less eachindex(A)`)

So we only need a  `getindex(A,i)` method and the `IndexStyle` trait.

But I am not sure whether doing row major iteration could not mess up generic code. E.g. you iterate over two matrices with linear indices, a Julia matrix and a Polymake matrix, this will be a mess.

sebasguts:
@saschatimme I will add the tests tomorrow, it might be worth it if we change the syntax again some time soon. I would guess these bugfixes in convert were necessary because of the last change we did, so lets try to not break it again.

kalmarek:
@saschatimme yeah, if You know exactly what to do, then no effort;)

and that's a good remark on joint iteration. I think it settles that row-major iteration is a bad idea;

sebasguts:
@kalmarek @saschatimme I added the tests.

kalmarek:
thanks!

PR #90, opened 2019-01-23
add installation from `Project.toml`

kalmarek:


kalmarek:
ok, it's not `Manifest.toml`, but `Project.toml`, but how do I change that, since the branch is protected?

sebasguts:
I fixed this. But isn't it easier to do `add  <url>`?

kalmarek:
I meant to add change in README.md, not the title of pull ;-)

`add url`  will add it to the global environment. 
`pull` + `activate` create a separate env, which you can safely delete, move afterwards

sebasguts:
You should provide PR's from your own fork ;).

I would nevertheless vote for `add` as the default version. We can still keep this as a second option for people who want to have separate env's. I guess most people (like me 28 hours ago) do not even know they exist.

PR #91, opened 2019-01-23
update notes on installation

kalmarek:


kalmarek:
so how about registering `Polymake.jl` in the blessed global repo, so that `add url` can be replaced by `add Polymake` ? @saschatimme ?

sebasguts:
How hard it is to get updates into that repo?

kalmarek:
> How hard it is to get updates into that repo?

@saschatimme says You make a release on github which is automatically referenced a few hours later

sebasguts:
So we should try to setup a travis that does releases for us? Or should we make them manually

PR #96, opened 2019-01-25
Hide properties in HTML by default

sebasguts:
@saschatimme wanted to add some CSS to make this look nicer


saschatimme:
By now I think we don't need any fancy CSS here.
I rebased the PR onto master and added some small string manipulation (removing the "type: " and "description: " prefixes).

saschatimme:
Screenshot 2019-04-04 at 10 38 47
Screenshot 2019-04-04 at 10 38 42

PR #97, opened 2019-01-25
Added keyword argument for typename params

sebasguts:


benlorenz:
We have published polymake 3.3 which contains the relevant changes on the polymake side and I will test this as soon as I have created new BinaryBuilder packages.

sebasguts:
Very nice, thank you :)

sebasguts:
@benlorenz Thanks. In which order do we need to merge #97 , #105 , and #106 ?

benlorenz:
Merging the update to 3.3 (#105) first would make most sense, for the other two pull requests the order should not matter.

Any other order would probably work as well since #106 is based on the update branch and I merged the update into #97 as well to make travis happy. If you prefer a cleaner history we could also rebase both branches after merging #105.

sebasguts:
Okay, will try to do it then. I usually prefer the rebase, but Github should be able to do it by itself

Issue #98, opened 2019-01-25
don't call `eval` in init


kalmarek:
when `using Polymake` inside a different module it produces:
```
WARNING: eval from module Polymake to OscarPolytope:    
Expr(:ref, :WrappedTypes, Expr(:call, :Symbol, "pm::perl::Object")) = Expr(:., :Polymake, :(:to_perl_object))
  ** incremental compilation may be broken for this module **
```
or similar.

The solution might be to populate (and save) `WrappedTypes` at build time.

PR #99, opened 2019-01-26
Removed eval from init

sebasguts:
Solves #98


kalmarek:
since we don't need all of those WrappedTypes for init that's ok to populate it at runtime

Issue #100, opened 2019-01-26
Add a "Rosetta Stone" to the README


sebasguts:
It would be nice to have a small list of comparisons between polymake and Polymake.jl code in the README, i.e., lines like
```
$x->VERTICES => x.VERTICES
```

kalmarek:
First draft:

# Rosetta stone:

### Variables

| Polymake                              | Julia                                                        |
| ------------------------------------- | ------------------------------------------------------------ |
| `$p` (reference to 'scalar' variable) | `p` (reference to any variable)                              |
| `print $p;`                           | `print(p)` or `println(p)` or `@show p`, or just `p` in REPL |
| `$i=5; $j=6;`                         | `i,j = 5,6` or `i=5; j=6` (`;` is needed for separation, <br />can be used to suppress return value in REPL) |
| `$s = $i + $j; print $s;`             | `s = i + j`                                                  |

### Arrays

| Polymake                                                 | Julia                                                        |
| -------------------------------------------------------- | ------------------------------------------------------------ |
| Linear containers with random access                     | Linear containers with random access + all the algebra attached      |
| `@A = ("a", "b", "c");`                                  | `A = ["a", "b", "c"]`                                        |
| `$first = $A[0];`<br />(`first` is equal to `a`)         | `first = A[1]`<br />(note the `1`-based indexing!)           |
| `@A2 = (3,1,4,2);`                                       | `A2 = [3,1,4,2]`                                             |
| `print sort(@A2);` <br />(a copy of `A2` is sorted)      | `println(sort(A2))` <br />(to sort in place use `sort!(A2))` |
| `$arr = new Array<Int>([3,2,5]);` <br />(a `C++` object) | `arr = Int[3,2,5]` <br />(although the type would have been inferred) |
| `$arr->[0] = 100;`<br />(assignment)                     | `arr[1] = 100` <br />(assignment; returns `100`)             |

### Dictionaries/Hash Tables

| Polymake                       | Julia                                                        |
| ------------------------------ | ------------------------------------------------------------ |
| `%h = ();`                     | `h = Dict()`<br />it is **MUCH** better to provide types, e.g.<br />`h = Dict{String, Int}()` |
| `$h{"zero"}=0; $h{"four"}=4;`  | `h["zero"] = 0; h["four"] = 4`<br />(call returns the value) |
| `print keys %h;`               | `@show keys(h)` (order is not specified)                     |
[...]

PR #104, opened 2019-02-11
New structure for app modules

### LINKED ISSUE #103: Names for polymake apps
I have just encountered a problem with the polymake application names:

Polymake.jl currently exports application names, so we can access functions, e.g., `polytope.cube`. Unfortunately, Polymake.jl exports those without importing those names beforehand, because they are modules, so importing those names does not make sense. This leads to a confusion if you use Hecke and Polymake, because both currently export `ideal`, with a complete different meaning. And I assume that this might cause problems with other OSCAR components soonish. So what should we do?
* Not export them?
* Rename them into something like `app_ideal`
* . . .

Ideas highly appreaciated!

sebasguts:
* App modules do not get exported in module Polymake anymore
* App modules export their functions

Closes #103

### DIFF
--- a/deps/parser/app_setup.jl
+++ b/deps/parser/app_setup.jl
@@ -1,0 +1,12 @@
+appname_dict = Dict(
+  :common  => :Common,
+  :fan  => :Fans,
+  :fulton  => :Fulton,
+  :graph  => :Graphs,
+  :group  => :Groups,
+  :ideal  => :Ideals,
+  :matroid  => :Matroids,
+  :polytope  => :Polytopes,
+  :topaz  => :Topaz,
+  :tropical  => :Tropical
+)
--- a/deps/parser/parser.jl
+++ b/deps/parser/parser.jl
@@ -33,6 +33,8 @@ function $julia_name(args...; template_parameters::Array{String,1}=String[], kee
     end
 end
 
+export $julia_name
+
 """
 end
 
@@ -52,6 +54,8 @@ function $julia_name(dispatch_object,args...; keep_PropertyValue=false, call_as_
         return convert_from_property_value(return_value)
     end
 end
+
+export $julia_name
 
 """
 end
--- a/deps/parser/parser.jl
+++ b/deps/parser/parser.jl
@@ -81,10 +81,15 @@ function parse_definition(method_dict::Dict, app_name::String, uniqueness_checke
 
 end
 
-function parse_app_definitions(filename::String,outputfileposition::String,include_file::String,additional_json_files_path::String)
+function parse_app_definitions(filename::String,outputfileposition::String,include_file::String,additional_json_files_path::String,appname_dict::Dict{Symbol,Symbol})
     println("Parsing "*filename)
     parsed_dict = JSON.Parser.parsefile(filename)
     app_name = parsed_dict["app"]
+    if !haskey(appname_dict,Symbol(app_name))
+        @warn "Application $app_name not found in translator"
+        appname_dict[Symbol(app_name)] = Symbol(app_name)
+    end
+    app_name_module = string(appname_dict[Symbol(app_name)])
     additional_json_file_name = abspath( joinpath( additional_json_files_path, app_name*".json" ) )
     if isfile(additional_json_file_name)
         additional_dict = JSON.Parser.parsefile(additional_json_file_name)
@@ -96,7 +101,7 @@ function parse_app_definitions(filename::String,outputfileposition::String,inclu
         end
     end
     return_string = """
-module $app_name
+module $app_name_module
 
 import ..internal_call_function, ..internal_call_method,
        ..internal_call_function_void, ..internal_call_method_void,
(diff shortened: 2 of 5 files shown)


saschatimme:
I like that the modules export their functions but I somewhat disagree that the modules aren’t exported anymore. The problems in #103 only occur when you load both modules with `using`. It is good practice to only use import in library level code and the using in the REPL. If there are name collisions in the REPL then you have to distinguish them by specifying the prefix. But with not exporting the modules we are basically always ending with the fallback method.

sebasguts:
Agreed that people should use `import` in their written code.

Normally I would agree that we should export it. The problem is that we cannot import `ideal`, because we create a new constant.

In the Julia world, how is it normally solved when two packages export the same name, without dependencies from each other? Would people simply be fine with the fact that they cannot write `using` for both? Or is there a way to define something like: "export, but only if not defined already"

kalmarek:
the standard practice is that You export functions, and if user wish to use two modules with the same exported name they have to specify the module.

sebasguts:
So they are forced to `import` the module instead of `using` it?

saschatimme:
You can still call `using` you just have to specify the modules for all collisions.

More general if two Packages want to define methods for the same functions, then they should have a common dependency on some package which just defines these functions. Then both can export those. But if there is a clash between module names and functions then this doesn't help (but usually in Julia modules are capitalized, so this does not happen). There are plenty of packages in the Julia exosystem with the name *Base.jl which define some functions and some default capabilities. And then other packages can depend on this to extend the functionality.

sebasguts:
This collision really only happens with modules, right? Types can have functions, too, right?

saschatimme:
You mean I have a type `Polytope` and somebody has a module `Polytope`? I think there is some convention in pluralizing modules, i.e., the modules would be `Polytopes` and it exports a type `Polytope`.

sebasguts:
No, I mean somebody has a `struct Polytope` and I have a function `Polytope`. That works I think. The problem is really just with modules.

saschatimme:
You should only have capitalized functions for constructors.

sebasguts:
But this is just a convention. Julia does not check this, so it is legal to do otherwise.

kalmarek:
@sebasguts Your `Polytope` function will just add methods to the existing `Polytope` constructor (which may or may not return an instance of `Polytope`, damn be the users;-)

i.e. that would happen if You intend to *extend* someones `Polytope` function and import it first. If You don't, then these have to be specified i.e. `MyModule.Polytope(...)` (for your function), `SomeonesModule.Polytope(...)` (for a constructor of the struct)

sebasguts:
Okay, I see. So we agree on still exporting the modules?

Should we still also export the functions from the modules, so one can `using Polymake.polytope`

kalmarek:
I'm for exporting modules, but I feel ambiguous about exporting functions from modules;

As far as I understand in polymake shell You choose an application (which corresponds to `using Polymake; using Polymake.polytope;`) and then You have direct access to all functions from that application. You may still access the other functions by prepending the app name, e.g. 
```
polytope > $s = simplex(3); print $s;
Polymake::polytope::Polytope__Rational=ARRAY(0x5574c9765230)
polytope > $s2 = topaz::simplex(3); print $s2;
Polymake::topaz::SimplicialComplex=ARRAY(0x5574c9765620)
```
and this will correspond to:
```
using Polymake; using Polymake.polytope;
s = simplex(3) # polytopal simplex, since polytope.simplex was brought into main
s2 = topaz.simplex(3) # simplicial-complex
```

kalmarek:
on the second thought, maybe we don't have to mimic polymake in this case?
but if we export all functions users will be faced with bunch of warnings when they `using Polymake`...?

sebasguts:
In a sense we seem to already have agreed that when people type `using foo`, they need to know what they are doing. So if they get the warnings, the cannot make the two using statements.

benlorenz:
In some way we are trying to please to different types of customers, those coming from a polymake background and those more familiar with julia. Maybe we should keep the corresponding names and exports somewhat separate:
We could rename our application modules according to julia standards, capitalized and some with plural: Common, Fans, Fulton, Graphs, Groups, Ideals, Matroids, Polytopes, Topaz and Tropical. These would be exported by default when 'using Polymake'.
Then we could add a second compatibility module (PolymakeCompat, PolymakeLingo, ...? Or is there some other way saying I want to load module X with feature Y?) for the polymake-users that exports the the lower case applications. We might also hide some of the more ugly hacks for polymake-style access to properties or functions there.

Also I think the applications should (in both variants) export their functions.

sebasguts:
I like @benlorenz suggestion, and will implement it now :)

kalmarek:
it went under my radar, i like the solution too

sebasguts:
I updated the PR to contain the following

* Functions from the appmodules are exported
* Appmodules are now capitalized and pluralized, which is done by a list. If there is an app which is not in the list, the module will get created with the normal name, and the parser gives an error.
* There is a module `Polymake.PolymakeCompat` that can be used or imported. It contains the translations to the standard polymake app names.

sebasguts:
For doing the `@pm` macro from #100 ? Indeed, that would be useful. I will autogenerate it in another PR, and make them actually included in the polymake module.

PR #105, opened 2019-02-18
update to polymake 3.3 binaries

benlorenz:
also use correct polymake_config path during json generation
also write binarybuilder deps file


benlorenz:
Dont merge yet, I need to do some rebuilds to fix an issue with automatic wrapper compilation.

PR #113, opened 2019-02-25
Make `cmake` executable configurable

sebasguts:


benlorenz:
I'm not really fond of adding more environment variables. Maybe we could depend on CMake.jl:
```julia
julia> import Pkg; Pkg.add("CMake");
...
julia> using CMake
[ Info: Precompiling CMake [631607c0-34d2-5d66-819e-eb0f9aa2061a]

julia> cmake
"/home/lorenz/.julia/packages/CMake/ZGawK/deps/usr/bin/cmake"
```

sebasguts:
Also a possibility. I was fixing the Polymake.jl installation on CentOS, and on CentOS the right cMake executable is `cmake3`. So I thought that would be an easy option. We can also depend on Cmake.jl, but I am not sure I like to have that somehow "unnecessary" dependeny.

benlorenz:
> We can also depend on Cmake.jl, but I am not sure I like to have that somehow "unnecessary" dependeny.

Julia itself is already a huge pile of unnecessary dependencies on most systems.

Maybe just use `cmake3` when it exists in the path and fall back to `cmake` otherwise.

sebasguts:
On the other hand, depending on CMake.jl would free us from having people to install cmake on their systems at all, so maybe it is the right option. I will prepare a PR for this.

sebasguts:
Resolved by #114

PR #117, opened 2019-02-25
@pm macro et al

kalmarek:
adds @pm macro, e.g.
`@pm Polytopes.Polytope{QuadraticExtension}(POINTS=...)`
constructs the polytope without `Polymake.jl` having to worry about `QuardaticExtension`s.

functionname has to fully specified;

@saschatimme please have a look at macro
@sebasguts this also changes the way WrappedTypes dictionary is filled (this has been sitting in my branch for quite some time).

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ The following tables explain by example how to quickly translate `Polymake` synt
 | `convert_to<Rational>($m_int)`                               | `convert(Diagonal{Rational{Int}}, m_int)`                    |
 | `$z_vec=zero_vector<Int>($m_int->rows)`<br />`$extended_matrix=($z_vec\|$m_int);`<br />(adds `z_vec` as the first column, result is dense) | `z_vec = zeros(Int, size(m_int, 1))`<br />`extended_matrix = hcat(z_vec, m_int)`<br />(result is sparse) |
 | `$set=new Set<Int>(3,2,5);`<br />`$template_Ex=new Array<Set<Int>>((new Set<Int>(5,2,6)),$set)` | `set = Set([3,2,5]); template_Ex = [Set([5,2,6]), set]`      |
-| `$p=new Polytope<Rational>(POINTS=>cube(4)->VERTICES);`<br />`$lp=new LinearProgram<Rational>(LINEAR_OBJECTIVE=>[0,1,1,1,1]);`<br />`$p->LP=$lp;`<br />`$p->LP->MAXIMAL_VALUE;` | `p = @pm Polytopes.Polytope(:POINTS=>Polymake.polytope.cube(4).VERTICES)`<br />`lp = @pm Polytopes.LinearProgram(:LINEAR_OBJECTIVE=>[0,1,1,1,1])`<br />`p.LP = lp`<br />`p.LP.MAXIMAL_VALUE` |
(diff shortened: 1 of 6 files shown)


kalmarek:
travis fails:
```
@pm(Polytopes.Polytope{QuadraticExtension}(input_dict_int)) isa pm_perl_Object
  C++/perl Interface module compilation failed; most likely due to a type mismatch.
  Set the variable $Polymake::User::Verbose::cpp to a positive value and repeat for more details.
```

locally `@pm(Polytopes.Polytope{QuadraticExtension}(input_dict_int))` works?
same error for `@pm(Tropical.Polytope{Max}(input_dict_int)) isa pm_perl_Object`,  `@pm(Tropical.Polytope{Max, Rational}(input_dict_int)) isa pm_perl_Object` and `@pm(Tropical.Polytope{Max, QuadraticExtension}(input_dict_int)) isa pm_perl_Object`.

then `polytope not defined` on line 80:
`:POINTS=>Polymake.polytope.cube(4).VERTICES`
@sebasguts didn't we defined those?

kalmarek:
ok, now only the constructors are failing; @benlorenz could You have a look?

sebasguts:
@kalmarek I quite like this approach.

What exactly do you change? You only use keyword arguments for `perlobj`?

sebasguts:
I quite like that syntactical approach, using the `@pm` macro. It is much more readable than the `perlobj` function. For interfacing from GAP, we still have the `perlobj`, so that is fine.

I will complete my review tomorrow.

kalmarek:
@sebasguts in perlobj every combination (kwarg, Pair{String,Sth}, Pair{Symbol, Sth}, Dict) is still possible;

`WrappedTypes` are filled with `fill_wrapped_types!(WrappedTypes, get_type_names())` instead of
`enhance_wrapped_type_dict()`;

sebasguts:
@kalmar The new translation dict seems unnecessary. We can just create it from the dict I already use to rename the application modules. It is just not really exported into polymake atm. I will provide a PR that does that, then we can rebase this one here. Would that be okay?

About the meta-module: Why not. Although most of it is generated anyway, so I do not really see that much necessity.

kalmarek:
@sebasguts you could just PR your changes to this branch @ my Polymake.jl repo

kalmarek:
@sebasguts and the point of meta-module is to replace generation of code as string as we do it right now, so that we could avoid things like
https://github.com/oscar-system/Polymake.jl/commit/986bcff19970d91463fe19623bf72b61a64c3a86
;-)

kalmarek:
```julia
function parse_function_call(expr)
    @assert expr.head == :call
    func = expr.args[1] # called function
    args = expr.args[2:end] # its arguments

    templates = Symbol[]
    # grab template parameters if present
    if func.head == :curly
        templates = func.args[2:end]
        func = func.args[1]
    end

    @assert func.head == Symbol('.')
    module_name = func.args[1]
    func_name = func.args[2].value

    return module_name, func_name, templates, args
end

macro pm_function(expr, void=false)
    app, polymake_func, templates, args = parse_function_call(expr)

    haskey(module_appname_dict, app) || throw("Module '$app' not registered in Polymake.jl.")
    polymake_app = module_appname_dict[app]
    polymake_func_name = qualified_polymake_name(polymake_app, polymake_func)

    if eval(void)
        quote
            internal_call_function_void($polymake_func_name, string.($templates),
                c_arguments($args...))
            return
        end
    else
        quote
            return_value = internal_call_function($polymake_func_name, string.($templates),
                    c_arguments($args...))
            return convert_from_property_value(return_value)
        end
    end
end
```
allows for
`Polymake.@pm_function Tropical.cyclic{Max}(3,5)` bypassing the generated code;
(and You can even `Polymake.@pm_function Tropical.cyclic{Max}(3,5) void=true` if You are not interested in the result)

sebasguts:
To me, this looks very good.

PR #119, opened 2019-03-05
Enh/new codegen

kalmarek:
Features:
* codegen happens at `using Polymake`, once; We could wrap the content of `applications.jl` in
```jula
let forced_generation=true
    (...)
    if forced_generation || (the other condition)
        (...)
   end
end
```
to make it at every pre-compile time;
* I took my liberty to make all methods (and functions) lowercase (see 102e1dd )
* You can call `@pm Polytopes.anyfunction{Its, Template, Parameters}(args...)`
* in @pm macro we detect function/BigObject constructor based on the case of the first letter :D any suggestions?
* I renamed `PolymakeCompat` module to simply `Compat`


kalmarek:
TODOs:
* get rid of the WARNING?
* BinaryBuilder doesn't seem to like `Compat` module (??)
* remove old parsing

src/functions.jl
```diff
@@ -102,7 +108,7 @@ Call a polymake method on the object `obj` with the given `func` name and given
 If `void=true` the function is called in a void context. For example this is important for visualization.
 """
 function call_method(obj, func::Symbol, args...; void=false, unwrap=true, kwargs...)
-    fname = string(func)
+    fname = uppercase(string(func))
```
sebasguts: I am not sure this actually works. For example `faces_of_dim` is a user method.

kalmarek: true, I'll fix that

src/meta.jl
```diff
@@ -0,0 +1,238 @@
...
+    end
+
+    callables = [PolymakeCallable(app_name,f) for f in app_json["functions"] if haskey(f, "name")]
+
+    callables = unique(pc -> pm_name(pc), callables)
+
+    return PolymakeApp(jl_module, app_name, callables)
+end
+
+function PolymakeApp(module_name::Symbol, json_file::String, module_file::String)
```
sebasguts: This method seems a bit weird. Both `module_file` and `module_name` seem to be unused. Is that on purpose?

kalmarek:
forgot to mention that docstrings for **functions** are available as well (of course with polymake syntax, should we do something about it??), e.g.

```julia
help?> Polytopes.cube
  cube<Scalar>(d; xup, xlow, Options) -> Polytope<Scalar>

  Produce a d-dimensional cube. Regular polytope corresponding to the Coxeter group of type B<sub>d-1</sub> = C<sub>d-1</sub>.

  The bounding hyperplanes are x<sub>i</sub> <= xup and x<sub>i</sub> >= xlow.

  Type Parameters: Scalar Coordinate type of the resulting polytope. Unless specified explicitly, deduced from the type of bound values, defaults to Rational.

  Arguments: Int d the dimension Scalar xup upper bound in each dimension Scalar xlow lower bound in each dimension

  Options: group => Bool add a symmetry group description to the resulting polytope character_table => Bool add the character table to the symmetry group description, if 0<d<7; default 1

  Returns Polytope<Scalar> 

  Examples:

  *) This yields a +/-1 cube of dimension 3 and stores it in the variable :c.

  │:c
  │
  │  = cube(3);

  *) This stores a standard unit cube of dimension 3 in the variable :c.

  │:c
  │
  │  = cube(3,0);

  *) This prints the area of a square with side length 4 translated to have its vertex barycenter at [5,5]:

  │  print cube(2,7,3)->VOLUME;

   16

```

sebasguts:
@kalmarek I do not think we can do anything about the polymake syntax. Doesn't that apply to basically anything from the Polymake doc's? Maybe we should print some warning, like
> This is in Polymake Syntax
> Please refer to the Polymake.jl Readme for a translation guide

sebasguts:
Very nice work indeed!!

PR #123, opened 2019-03-14
Structure readme

saschatimme:
This adds an index to the readme which links to the subsections. Also the current state is moved to the end and Rosetta Stone is renamed to simply `Polymake syntax translation`.


sebasguts:
We should also overhaul the text about functions.

PR #125, opened 2019-04-01
Enh/rename modules

kalmarek:
as per discussion with Sascha and Michael;

I updated readme as well.


kalmarek:
yes, just capitalization, since we don't intend to have `Polytope` type on this level;

Although pluralisation would be way cooler with `Fultons` and `Tropics` ;-)

sebasguts:
@kalmarek feel free to merge it, or rebase it (needs to be done manually)

PR #135, opened 2019-04-30
update polymake binaries to 3.4

benlorenz:
also some more version bumps and switched to julia Yggdrasil for gmp/mpfr


benlorenz:
> Could you rebase this before we merge it?

done

PR #139, opened 2019-04-30
Cleanup wrappers when rebuilding polymake

### LINKED ISSUE #136: Clean polymake_user directory after polymake upgrades
This should avoid https://github.com/oscar-system/Polymake.jl/pull/134#issuecomment-487974262. It's just not clear to me what we should delete and what not.

sebasguts:
Closes #136

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -96,14 +96,20 @@ If you already have a polymake installation you need to set the environment vari
         using Pkg: depots1
         function prepare_env()
             ENV["PERL5LIB"]=$perllib
-            ENV["POLYMAKE_USER_DIR"] = abspath(joinpath(depots1(),"polymake_user"))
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -96,14 +96,20 @@ If you already have a polymake installation you need to set the environment vari
         using Pkg: depots1
         function prepare_env()
             ENV["PERL5LIB"]=$perllib
-            ENV["POLYMAKE_USER_DIR"] = abspath(joinpath(depots1(),"polymake_user"))
+            user_dir = ENV["POLYMAKE_USER_DIR"] = abspath(joinpath(depots1(),"polymake_user"))
+            if Base.Filesystem.isdir(user_dir)
+                del = filter(i -> Base.Filesystem.is_dir(i) && startswith(i, "wrappers."), readdir(user_dir))
```
saschatimme: This is probably `isdir` and not `is_dir`

sebasguts: Indeed. Fixed.

PR #143, opened 2019-05-13
more deps via binarybuilder

benlorenz:
new build with more dependencies via binarybuilder, also with bliss and ppl now

lets see what travis says ....


sebasguts:
Very nice, thank you :)

PR #145, opened 2019-05-28
Fix/convert sets

kalmarek:


kalmarek:
hmm... I can't reproduce the fail locally

kalmarek:
I could reproduce locally, but only on julia-1.0.4. (1.1 runs fine, contrary to travis) lets see what happens now

kalmarek:
ok, this times it seems like it's passing both locally and on travis, however one needs to set `OMP_NUM_THREADS=1`;
@sebasguts should we `if ENV["OMP_NUM_THREADS"] == 1` the added test?

kalmarek:
@sebasguts please have a look, it's ready

test/perlobj.jl
```diff
@@ -150,8 +150,10 @@
 
         p.MILP = @pm Polytope.MixedIntegerLinearProgram( LINEAR_OBJECTIVE = obj, INTEGER_VARIABLES = intvar)
 
-        # segfaults when called from different thread, see
-        # https://github.com/oscar-system/Polymake.jl/issues/144
-        @test p.MILP.MINIMAL_VALUE == -7
+        if ENV["OMP_NUM_THREADS"] == 1
```
kalmarek: of course this will fail when `OMP_NUM_THREADS` is not defined, we need to `get(ENV, "OMP_NUM_THREADS", 1)` it

thanks @saschatimme for thorough review :D (no offence, I noted this only late in the evening :P)

PR #146, opened 2019-06-12
remove superfluous return from @pm macro

kalmarek:
the return statement was causing returns from functions in "unexpected" places (i.e. right after application of `@pm` :P

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -183,7 +183,7 @@ POINTS
 
 
 ```
-Note: the expression in `@pm` macro is parsed syntactically, so it has to be a valid `julia` expression. However template parameters NEED NOT to be defined in `julia`, but must be valid names of `polymake` property types.
+Note: the expression in `@pm` macro is parsed syntactically, so it has to be a valid `julia` expression. However template parameters NEED NOT to be defined in `julia`, but must be valid names of `polymake` property types. Nested types (such as `{QuadraticExtension{Rational}}`) are allowed.
 """
 macro pm(expr)
     module_name, polymake_func, templates, args, kwargs = Meta.parse_function_call(expr)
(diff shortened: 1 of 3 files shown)


src/functions.jl
```diff
@@ -139,6 +139,52 @@ end
...
 
+
+"""
+    @pm PolymakeModule.function_name{Template, parameters}(args)
+
+This macro can be used to
+ * create `polymake` Big Objects (such as polytopes)
+ * call `polymake` functions with specific template parameters.
+
+The expression passed to the macro has to be fully qualified name (starting with the uppercase name of polymake application) of a `polymake` object or function, with template parameters enclosed in `{ ... }`.
```
saschatimme:
```suggestion
The expression passed to the macro has to be the fully qualified name (starting with the uppercase name of polymake application) of a `polymake` object or function, with template parameters enclosed in `{ ... }`.
```

src/functions.jl
```diff
@@ -139,6 +139,52 @@ end
...
+type: Polytope<Max, Rational>
+
+POINTS
+0 -1 -1
+0 1 0
+
+
+
+```
+Note: the expression in `@pm` macro is parsed syntactically, so it has to be a valid `julia` expression. However template parameters NEED NOT to be defined in `julia`, but must be valid names of `polymake` property types. Nested types (such as `{QuadraticExtension{Rational}}`) are allowed.
```
saschatimme:
```suggestion
Note: the expression in `@pm` macro is parsed syntactically, so it has to be a valid `julia` expression. However template parameters **need not** to be defined in `julia`, but must be valid names of `polymake` property types. Nested types (such as `{QuadraticExtension{Rational}}`) are allowed.
```

kalmarek:
> Looks good just two small things (to show that I do thorough reviews :P)

ah, it's so good to have You around ;-)

PR #148, opened 2019-06-13
Fix/converts take two

kalmarek:
I hope there won't be take three in the near future

### DIFF
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -37,9 +37,9 @@ function Base.append!(A::pm_Array{T}, itr) where T
 end
 
 # workarounds for pm_Array{String}
-pm_Array{T}(n::Integer) where T<:AbstractString = pm_Array{AbstractString}(convert(Int64, n))
+pm_Array{T}(n::Integer) where T<:AbstractString = pm_Array{AbstractString}(Int64(n))
 pm_Array{T}(n::Integer, elt::T) where T<:AbstractString =
-pm_Array{AbstractString}(convert(Int64, n), elt)
+pm_Array{AbstractString}(Int64(n), elt)
 
 Base.@propagate_inbounds function Base.setindex!(A::pm_Array{S}, val, n::Integer) where {S<:AbstractString}
     @boundscheck 1 <= n <= length(A) || throw(BoundsError(A, n))
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -1,10 +1,12 @@
-function pm_Array{T}(vec::AbstractVector) where T
+const pm_Array_suppT = Union{Int32, Int64, String, pm_Set{Int32}, pm_Matrix{pm_Integer}, AbstractString}
+
+function pm_Array{T}(vec::AbstractVector) where T <: pm_Array_suppT
     arr = pm_Array{T}(length(vec))
     @inbounds arr .= vec
     return arr
 end
 
-pm_Array{T}(n::Integer, elt) where T = pm_Array{T}(Int64(n), T(elt))
+pm_Array{T}(n::Integer, elt) where T<:pm_Array_suppT = pm_Array{T}(Int64(n), T(elt))
 pm_Array(n::Integer, elt::T) where T = pm_Array{T}(Int64(n), elt)
 
 pm_Array(vec::AbstractVector) = pm_Array{convert_to_pm_type(eltype(vec))}(vec)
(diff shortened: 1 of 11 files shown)


src/arrays.jl
```diff
@@ -1,10 +1,12 @@
-function pm_Array{T}(vec::AbstractVector) where T
+const pm_Array_suppT = Union{Int32, Int64, String, pm_Set{Int32}, pm_Matrix{pm_Integer}, AbstractString}
```
saschatimme: This has to be changed for each new small type we bind, right? This looks like it will introduce bugs at some point :)

src/convert.jl
```diff
@@ -1,43 +1,38 @@
...
-convert(::Type{pm_Set}, itr) = pm_Set(itr)
-convert(::Type{pm_Set{T}}, itr) where T = pm_Set{T}(itr)
-convert(::Type{pm_Set{T}}, as::AbstractSet) where T = pm_Set{T}(as)
+####################  Guessing the polymake type  ######################
 
-# disambiguations:
-convert(::Type{pm_Set}, as::AbstractSet) = pm_Set(as)
-convert(::Type{pm_Set{T}}, s::pm_Set{T}) where T = s
+# By default we throw an error:
+convert_to_pm_type(T::Type) = throw(ArgumentError("Unrecognized argument type: $T.\nYou need to convert to polymake compatible type first."))
```
saschatimme: Is this intended to be a public function?

src/matrices.jl
```diff
@@ -1,15 +1,24 @@
...
+
+@inline function pm_Matrix{T}(mat::AbstractMatrix) where T <: pm_Matrix_suppT
     res = pm_Matrix{T}(size(mat)...)
     @inbounds res .= mat
     return res
 end
 
 # needed for broadcast
-@inline function pm_Matrix{T}(rows::UR, cols::UR) where {T, UR<:Base.AbstractUnitRange}
+@inline function pm_Matrix{T}(rows::UR, cols::UR) where {T <: pm_Matrix_suppT, UR<:Base.AbstractUnitRange}
```
saschatimme: Q: Can we only participate with unit ranges?

src/sets.jl
```diff
@@ -1,16 +1,14 @@
 export incl, swap
 
-import Base: Set
+const pm_Set_suppT = Union{Int32, Int64}
```
saschatimme: Wasn't there set of sets at some point?

src/arrays.jl
kalmarek: this is just to make the signatures below shorter and to have just one place to be updated once a new element type is added. What do You suggest

src/convert.jl
kalmarek: no, external users should overload `convert(::Type{PolymakeType}, x::T) where T = ...` where the right hand side should consist of `pm_types`, or machine types.

src/matrices.jl
kalmarek: not sure what do you mean by participate? this is needed for similar, which in turn is needed for broadcasting

src/sets.jl
kalmarek: no, there are `pm_Array`s of `pm_Set`s of `int`s; `pm_Set`s are only of `int` and `(long) long` ;)

src/matrices.jl
saschatimme: But what about broadcasting with non unit ranges, i.e., `1:2:9`? (But maybe not that important)

src/sets.jl
saschatimme: ah okay :)

src/arrays.jl
saschatimme: I thought these should match the types listed [here](https://github.com/oscar-system/Polymake.jl/blob/6ae6fb7026debab304db837c6b6a51b32c37d7ee/deps/type_setup.jl#L14-L23) or is this just for a subset?

kalmarek: that's right. Those are wrapped, but never used (and obviously never tested).
We could add support on to construct them on the julia side if we wish (you can interact with them if polymake returns them, just fine), but I think it's not particularly necessary (until some voices its demand).

And no one is willing to add tests for them anyway :P

src/matrices.jl
kalmarek: I added this a while ago when I wanted to make `f.(::pm_Matrix)` return a `pm_Matrix` instead of `Array{..., 2}`. It seems that we don't need non-unital ranges for this.

kalmarek:
@saschatimme please resolve conversations and if everything else looks good hit merge ;-) this has been hanging here for more than a month now...

PR #156, opened 2019-09-11
return Arrays for unknown types

kalmarek:


src/matrices.jl
```diff
@@ -39,10 +39,18 @@ function Base.similar(mat::pm_Matrix, ::Type{S}, dims::Dims{2}) where S<:Union{p
     return pm_Matrix{convert_to_pm_type(S)}(dims...)
 end
 
-function Base.similar(mat::pm_Matrix, ::Type{S}, dims::Dims{1}) where S<:Union{pm_Integer, pm_Rational, Float64}
+function Base.similar(mat::pm_Matrix, ::Type{S}, dims::Dims{2}) where S
+    return Matrix{S}(undef, dims...)
+end
+
+function Base.similar(mat::pm_Matrix, ::Type{S}, dims::Dims{1}) where S<:Union{pm_Integer, pm_Rational}
```
saschatimme: Can you remove the `Float64` type in the above definition for `Dims{2}` to keep the symmetry in the definitions?

README.md
```diff
@@ -2,32 +2,39 @@
 
 [![Build Status](https://travis-ci.com/oscar-system/Polymake.jl.svg?branch=master)](https://travis-ci.com/oscar-system/Polymake.jl)
 
-Julia package for using [`polymake`](https://polymake.org/doku.php), a software for research in polyhedral geometry.
+`Polymake.jl` is a julia package for using [`polymake`](https://polymake.org/doku.php), a software for research in polyhedral geometry from julia.
```
saschatimme:
```suggestion
`Polymake.jl` is a Julia package for using [`polymake`](https://polymake.org/doku.php), a software for research in polyhedral geometry from Julia.
```

Julia is always capitalized I think

saschatimme: I don't mark the other occurrences :)

README.md
```diff
@@ -2,32 +2,39 @@
...
 
-The installation can be done in the Julia REPL by first pressing `]` and then
+The installation can be done in the julia REPL by executing
 ```julia
-pkg> add Polymake
+julia> using Pkg; Pkg.add("Polymake")
 ```
 This will fetch a pre-build binary of `polymake`. You are ready to start `using Polymake`.
 
+Note: Pre-build binaries are available only for the `Linux` platform. MacOSX users need to follow the "Your own installation of `polymake`" below. Windows users are encouraged to try running `julia` inside Window Subsystem for Linux and reporting back ;)
```
saschatimme: Its either `macOS` (new name) or `OS X` (old name)

README.md
```diff
@@ -2,32 +2,39 @@
...
+
 ### Your own installation of `polymake`
 
-If you already have a recent enough version of `polymake` on your system, you either need to make `polymake-config` available in your `PATH` or the environment variable `POLYMAKE_CONFIG` needs to point to the correct `polymake-config` file.
+If you already have a recent enough version of `polymake` on your system,
```
saschatimme: Add the concrete polymake version here. 3.2? 3.3?

README.md
```diff
@@ -79,21 +88,123 @@ POINTS
...
+### Example script
+
+The following script is modelled on the one from the [Using Perl within polymake](https://polymake.org/doku.php/user_guide/tutorials/perl_intro) tutorial:
+
+```julia
+using Polymake
+
+str = read("points.demo", String)
+matrix_str = "["*replace(str, "/"=>"//")*"]"
+matrix = eval(Base.Meta.parse(matrix_str))
```
saschatimme: Do we really need an `eval` here? Or maybe we should provide a helper function to hide this?

README.md
```diff
@@ -79,21 +88,123 @@ POINTS
...
+@show matrix
+
+p = @pm Polytope.Polytope(:POINTS=>matrix)
```
saschatimme: I think we should aim for a favourite constructor (since we support all kind of variants).
I like `@pm Polytope.Polytope(POINTS = matrix)` more since we don't need to explain symbols and it is more consistent with kwargs constructors.

README.md
```diff
@@ -64,7 +71,9 @@ Just remember that You need to `activate Polymake.jl` to use `Polymake`.
 
 ## Examples
 
-`polymake` big objects (like `Polytope`, `Cone`, etc) can be created with the `@pm` macro:
+In this section we just highlight various possible uses of `julia/polymake`. Please refer to [Polymake syntax translation](#polymake-syntax-translation) for more thorough treatment.
```
saschatimme: I like `Polymake.jl` more than `julia/polymake`. And `polymake` instead of `perl/polymake`.

src/matrices.jl
kalmarek: If `M` is a `pm_Matrix{pm_Rational}` slicing `M[1, :]` results in `pm_Vector{pm_Rational}`, and `M[[1], :]` returns a `pm_Matrix{pm_Rational}` (as expected). However if `M` is a `pm_Matrix{Float64}` the latter works, but the former fails, since we don't wrap `pm_Vector{Float64}`. This makes `M[1, :]` return `Vector{Float64}` which is the best we can afford (at least until someone wraps `pm_Vector{Float64}`)

README.md
kalmarek: `3.3`, but it's indicated in the third line of README, should I repeat this here?

saschatimme: Cannot hurt

saschatimme:
We also get a weird error on 1.0 on travis. I have never seen this so far. Any idea?

README.md
kalmarek: I don't know actually what is the best way to do so, but this one was the shortest (e.g. there is no `Base.parse(::Type{Rational{T}, ...)`),
```
matrix = Meta.parse(matrix_str) |> eval
```
??? (more obscure, I'd say)

saschatimme: Yeah I thought one can use `DelimitedFiles` but this is not the case :(

README.md
kalmarek: I'm in favour of `kwargs` convention as well

README.md
saschatimme: Maybe just a comment that this is a hack?^^

PR #159, opened 2019-09-18
v0.2.1

saschatimme:
Let's make a new release with the two bug fixes #158 and #156


saschatimme:
Always a pleasure to work with such an amazing team!

PR #161, opened 2019-09-24
update readme for macOS binaries

benlorenz:
- corrected link to polymake repo (default releases branch is fine and more recent than snapshots)
- also mention POLYMAKE_CONFIG=no option to force binaries


saschatimme:
Do we also want to make a 0.2.2 release then?

benlorenz:
> Do we also want to make a 0.2.2 release then?

yes, or maybe wait a few days until we did a few more tests (if we can find more macs)

saschatimme:
Sure.

PR #170, opened 2019-11-15
Fix mismatch between pm_array cpp and pm_Array_suppT

saschatimme:
I noticed that we declared much less `pm_Array` element types on the Julia site than on the cpp side.

Who do I have to bribe to not write all the missing tests? 😅


saschatimme:
@benlorenz  I also found that `pm_Array{pm_Rational}` was missing, so I added this. And I also added some lightweight tests (things can be constructed at least...)

kalmarek:
oh no, so much new types, so little testing :P

Issue #177, opened 2019-12-05
BigObject name completion and help strings


kalmarek:
It'd be great to have a `json` file generated at compile time with a (potentially incomplete) list of big objects together with their docstrings. Then we can add annotated stub structs when we generate code for applications:

```julia
struct PolymakeObject <: PolymakeCallable
    jl_function::Symbol
    pm_name::String
    app_name::String
    json::Dict{String, Any}

   PolymakeObject(pm_name::String, app_name::String) = new(Symbol(pm_name), pm_name, app_name)
   PolymakeObject(jl_name::Symbol, pm_name::String, app_name::String, json_dict) = new(jl_name, pm_name, app_name, json_dict)
end

docstring(obj::PolymakeObject) = obj.json["docs"] # or whatever

function jl_code(obj::PolymakeObject)
    return :(
        struct $(jl_symbol(obj)) end;
        Base.Docs.getdoc(::typeof($(jl_symbol(obj)))) = Markdown.parse($(docstring(obj)))
    )
end
```

saschatimme:
Could we also use this to hook into the REPL auto complete to suggest BigObjects?

kalmarek:
that's the sole purpose, but maybe we can do even more...

kalmarek:
fixed in https://github.com/oscar-system/Polymake.jl/pull/178

PR #178, opened 2019-12-05
add objects with help and params to json

benlorenz:
first step for #177


benlorenz:
the additional json section is as follows:
```json
  "objects" : [
    {
      "help" : " A polyhedral cone, not necessarily pointed.\n Note that in contrast to the vertices of a polytope, the \u001b[4mRAYS\u001b[24m are given in affine coordinates.\n\nType Parameters:\n  \u001b[4mScalar\u001b[24m numeric data type used for the coordinates, must be an ordered field. Default is \u001b[4mRational\u001b[24m.\n",
      "name" : "Cone",
      "params" : [
        "Scalar"
      ]
    },
```

kalmarek:
```julia
julia> using Polymake
[ Info: Generating module Common
[ Info: Generating module Ideals
[ Info: Generating module Graphs
[ Info: Generating module Fulton
[ Info: Generating module Fans
[ Info: Generating module Groups
[ Info: Generating module Polytopes
[ Info: Generating module Topaz
[ Info: Generating module Tropical
[ Info: Generating module Matroids
polymake version 3.6
Copyright (c) 1997-2019
https://polymake.org

julia> Polytopes.Polytope{pm_Rational}(POINTS=[1 1 0; 1 4 1])
type: Polytope<Rational>

POINTS
1 1 0
1 4 1



julia> Tropical.Polytope{min, Rational}(POINTS=[1 1 0; 2 4 1])
type: Polytope<Min, Rational>

POINTS
0 0 -1
0 2 -1

```
this is just the first try;  I needed to pluralize certain application names, to avoid e.g. `Polytope.Polytope` (julia does not allow this). Interestingly this happened only, when plural does makes sense. Although I see great potential in application `Tropics` :) 

Comments welcome. @saschatimme @benlorenz @micjoswig 

btw, there is some excessive spacing when printing (and printing super large objects is by no means useful). Can we make something about it? does polymake support ellipsing in any sense (output filter?)

src/meta.jl
```diff
@@ -1,30 +1,51 @@
...
+    return "$app_name::$func_name<$(join(templates, ","))>"
+end
+
+translate_type_to_pm_string(::Type{Bool}) = "bool"
+translate_type_to_pm_string(::Type{Int32}) = "int"
+translate_type_to_pm_string(::Type{<:AbstractFloat}) = "Float"
+translate_type_to_pm_string(::Type{<:Rational}) = "Rational"
+translate_type_to_pm_string(::Type{<:pm_Rational}) = "Rational"
+translate_type_to_pm_string(::Type{<:Integer}) = "Integer"
+translate_type_to_pm_string(::typeof(min)) = "Min"
```
saschatimme: I think it would be better to define structs 'Min' and 'Max' instead of overloading the min and max functions.

kalmarek: @alexej-jordan has already implemented those for TropicalNumbers;
It's just a matter of convenience for the user; 
we can provide both overloads for `Type{pm_Min}` and `typeof(min)` so that both
```
Tropical.Polytope{pm_Min, Rational}(...)
Tropical.Polytope{min, Rational}(...)
```
work (I don't have strong preference for any of them).

kalmarek: btw. both constructors `Tropical.Polytope{min}(...)` and `Tropical.Polytope{min, pm_Rational}` are defined

micjoswig:
This is very nice, indeed.

Ellipsing is not supported yet.  However, we do support output filters.  So providing
one should not be a big deal.  Let's discuss this on Wednesday.

-- Michael 

On 06.12.19 07:51, kalmarek wrote:
> this is just the first try;  I needed to pluralize certain application names, to avoid e.g. `Polytope.Polytope` (julia does not allow this). Interestingly this happened only, when plural does makes sense. Although I see great potential in application `Tropics` :) 
> 
> Comments welcome. @saschatimme @benlorenz @micjoswig 
> 
> btw, there is some excessive spacing when printing (and printing super large objects is by no means useful). Can we make something about it? does polymake support ellipsing in any sense (output filter?)
> 

-- 
Prof. Dr. Michael Joswig           <joswig@math.tu-berlin.de>
Technische Universitaet      Institut fuer Mathematik, MA 6-2
Str. des 17. Juni 136                 D-10623 Berlin, Germany
phone +49 (30) 314-75904               fax +49 (30) 314-25047

src/app_setup.jl
```diff
@@ -1,14 +1,14 @@
...
+  :fan        => :Fans,
+  :fulton     => :Fulton,
+  :graph      => :Graphs,
+  :group      => :Groups,
+  :ideal      => :Ideals,
+  :matroid    => :Matroids,
+  :polytope   => :Polytopes,
+  :topaz      => :Topaz,
+  :tropical   => :Tropical
 )
```
saschatimme: Why do we need to rename the modules?

kalmarek:
```julia
julia> module Polytope
       struct Polytope end
       end
ERROR: invalid redefinition of constant Polytope
Stacktrace:
 [1] top-level scope at REPL[1]:2

```

saschatimme: ah yes. We already had this problem discussion ....

btw this would work 
```
julia> module polytope
           struct Polytope end
       end
Main.polytope
```

(but let's discuss this in person)

PR #179, opened 2019-12-10
add solve_LP

kalmarek:
`solve_LP` in `polymake_direct_calls` is an example on how to call directly `libpolymake` functions;

in principle  generating such functions should be possible without (much) human intervention

### DIFF
--- a/deps/src/polymake.cpp
+++ b/deps/src/polymake.cpp
@@ -20,6 +20,8 @@
 
 #include "polymake_caller.h"
 
+#include "polymake_direct_calls.h"
+
 #include "polymake_type_translations.h"
 
 #include "generated/type_declarations.h"
@@ -41,6 +43,8 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
     polymake_module_add_set(polymake);
 
     polymake_module_add_array(polymake);
+
+    polymake_module_add_direct_calls(polymake);
 
     polymake.method("initialize_polymake", &initialize_polymake);
     polymake.method("application", [](const std::string x) {
--- a/deps/src/polymake.cpp
+++ b/deps/src/polymake.cpp
@@ -48,16 +48,7 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
     });
 
     polymake.method("shell_execute", [](const std::string x) {
-        // FIXME: tuples with strings are broken in cxxwrap
-        // return res;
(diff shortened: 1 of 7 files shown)


deps/src/polymake.cpp
```diff
@@ -75,10 +70,7 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
         bool full=false, bool html=false){
         std::vector<std::string> ctx_help =
             data.main_polymake_session->shell_context_help(input, pos, full, html);
-        jl_value_t** doc_strings = new jl_value_t*[ctx_help.size()];
-        for (int i=0; i < ctx_help.size(); ++i)
-            doc_strings[i] = jl_cstr_to_string(ctx_help[i].c_str());
-        return jlcxx::make_julia_array(doc_strings, ctx_help.size());
+        return jlcxx::make_julia_array(&ctx_help[0], ctx_help.size());
```
saschatimme: Why is it no more necessary to convert from c-string to string?

deps/src/polymake.cpp
```diff
@@ -42,22 +44,15 @@ JLCXX_MODULE define_module_polymake(jlcxx::Module& polymake)
...
-        // return res;
-        // instead we return an array of a bool and three strings now
-        auto         res = data.main_polymake_session->shell_execute(x);
-        jl_value_t** output = new jl_value_t*[4];
-        output[0] = jl_box_bool(std::get<0>(res));
-        output[1] = jl_cstr_to_string(std::get<1>(res).c_str());
-        output[2] = jl_cstr_to_string(std::get<2>(res).c_str());
-        output[3] = jl_cstr_to_string(std::get<3>(res).c_str());
-        return jlcxx::make_julia_array(output, 4);
+        return data.main_polymake_session->shell_execute(x);
```
saschatimme: This is now no more broken in cxxwrap?

deps/src/polymake.cpp
kalmarek: I don't know why this was necessary in the first place;
Example here: https://github.com/JuliaInterop/libcxxwrap-julia/commit/b0068f102b190ece10bdfc9f7ac096c4d2ceefd0
shows that this is not needed

deps/src/polymake.cpp
kalmarek: I believe support for tuples (with pointers) arrived here:
https://github.com/JuliaInterop/CxxWrap.jl/issues/133

PR #181, opened 2019-12-21
Types/sparsematrix

alexej-jordan:
pm_SparseMatrix working; two exceptions:
1) When applying julia integer scalar operations (i.e. Int64) on an integer templated polymake sparse matrix we receive an Int64 sparse matrix, so most probably this case gets not caught by some broadcasting rule
2) Even though @kalmarek helped me to overload the way julia detects the element type when broadcasting a julia sparse matrix some special case still gives us an "Any"-typed julia sparse matrix. For pm_SparseMatrix this only affects the tests, but the compatibility of julia sparse matrices and our polymake number types is not guaranteed


src/broadcast.jl
```diff
@@ -12,8 +20,14 @@ Base.similar(X::Union{pm_Vector, pm_Matrix}, ::Type{S}, dims::Dims{2}) where
 
 Base.similar(X::Union{pm_Vector, pm_Matrix}, ::Type{S}, dims::Dims{2}) where {S} = Matrix{S}(dims...)
 
+Base.similar(X::pm_SparseMatrix, ::Type{S}, dims::Dims{2}) where
+    {S <: pm_VecOrMat_eltypes} = pm_SparseMatrix{convert_to_pm_type(S)}(dims...)
+
+Base.similar(X::pm_SparseMatrix, ::Type{S}, dims::Dims{2}) where {S} = SparseMatrixCSC{S}(dims...)
```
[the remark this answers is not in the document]
alexej-jordan: wouldn't that mean that broadcasting a `pm_SparseMatrix` gives us a `pm_Matrix` or a julia matrix? running the test script with this alteration yields two additional fails because we receive a dense matrix instead of a sparse one

alexej-jordan:
Addressed the requested changes.
Only missing thing that comes to my mind is to replace Broadcast.ArrayStyle with SparseArrays.HigherOrderFns.SparseMatStyle in the broadcast.jl file (line 43), but I obviously do not understand how to make this work. Even though, leaving it the way it is does not result in any unwanted behaviour, but probably this is more or less related to the already mentioned problem when broadcasting julia sparse matrices
@kalmarek 
@saschatimme

src/sparsematrix.jl
```diff
@@ -0,0 +1,72 @@
...
+end
+
+function SparseArrays.findnz(mat::pm_SparseMatrix{T}) where T <: pm_VecOrMat_eltypes
+    nzi = nzindices(mat)
+    ri = zeros(Int64,0)
+    ci = zeros(Int64,0)
+    v = zeros(T,0)
+    for r=1:length(nzi)
+        for c in nzi[r]
+            append!(ri,r)
```
kalmarek: if we know the size of `ri` beforehand, It's better to allocate first `ri = Vector{Int}(undef, len)` and then use additional variable `k` to , `ri[k] = r; k+=1`

src/sparsematrix.jl
```diff
@@ -0,0 +1,72 @@
...
+pm_SparseMatrix(mat::AbstractMatrix{T}) where T <: Union{Rational, pm_Rational} = pm_SparseMatrix{pm_Rational}(mat)
+pm_SparseMatrix(mat::AbstractMatrix{T}) where T <: AbstractFloat = pm_SparseMatrix{Float64}(mat)
+
+pm_SparseMatrix{T}(x...) where T <: Integer = pm_SparseMatrix{pm_Integer}(x...)
+pm_SparseMatrix{T}(x...) where T <: Rational = pm_SparseMatrix{pm_Rational}(x...)
+pm_SparseMatrix{T}(x...) where T <: AbstractFloat = pm_SparseMatrix{Float64}(x...)
+
+pm_SparseMatrix(mat::M) where M <: pm_SparseMatrix{Float64} = mat
+
+Base.size(m::pm_SparseMatrix) = (Int(rows(m)), Int(cols(m)))
```
kalmarek: add the same for `pm_Matrix`

alexej-jordan:
adjusted everything according to the requested changes. also, added a small randomized test for the findnz method

src/sparsematrix.jl
```diff
@@ -58,15 +54,29 @@ end
...
-            append!(ri,r)
-            append!(ci,c+1)
-            append!(v,mat[r,c+1])
+            ri[k] = r
+            ci[k] = c + 1
+            v[k] = mat[r, c + 1]
+            k += 1
         end
     end
+
```
kalmarek: remove also this part ;-)

kalmarek:
ok, other than casts to `Int`s in size of `pm_Matrix` and `pm_Vector` we're only waiting for @saschatimme comment on the broadcast issue.

saschatimme:
I do not have a good solution for the broadcasting problem, but I also don't think this is crucial for now.

I would say let's rebase and then merge. @kalmarek

PR #182, opened 2020-01-02
Types/tropicalnumber

alexej-jordan:
Working version of pm_TropicalNumber, together with the corresponding tests


alexej-jordan:
@kalmarek

src/tropicalnumber.jl
```diff
@@ -0,0 +1,44 @@
...
+
+function Base.:*(x::pm_TropicalNumber{A1,S1},y::pm_TropicalNumber{A2,S2}) where {A1,A2 <: pm_TropicalNumber_suppAddition, S1,S2 <: pm_TropicalNumber_suppScalar}
+    (A1 == A2) & (S1 == S2) || throw(ArgumentError("multiplication of two instances of pm_TropicalNumber only allowed when both parameter types match"))
+    return _mult(x,y)
+end
+
+function Base.://(x::pm_TropicalNumber{A1,S1},y::pm_TropicalNumber{A2,S2}) where {A1,A2 <: pm_TropicalNumber_suppAddition, S1,S2 <: pm_TropicalNumber_suppScalar}
+    (A1 == A2) & (S1 == S2) || throw(ArgumentError("division of two instances of pm_TropicalNumber only allowed when both parameter types match"))
+    return _div(x,y)
+end
```
kalmarek:
```
Base.:/(x::pm_TropicalNumber{A1,R}, y::pm_TropicalNumber{A2, R}) where{A1, A2, R <: pm_Rational} = x//y
```

src/tropicalnumber.jl
```diff
@@ -0,0 +1,44 @@
...
+Base.one(i::pm_TropicalNumber{A}) where A <: pm_TropicalNumber_suppAddition = pm_TropicalNumber{A}(0)
+Base.one(i::Type{<:pm_TropicalNumber{A}}) where A <: pm_TropicalNumber_suppAddition = pm_TropicalNumber{A}(0)
+dual_zero(i::pm_TropicalNumber{pm_Min}) = zero(pm_TropicalNumber{pm_Max})
+dual_zero(i::Type{<:pm_TropicalNumber{pm_Min}}) = zero(pm_TropicalNumber{pm_Max})
+dual_zero(i::pm_TropicalNumber{pm_Max}) = zero(pm_TropicalNumber{pm_Min})
+dual_zero(i::Type{<:pm_TropicalNumber{pm_Max}}) = zero(pm_TropicalNumber{pm_Min})
+orientation(i::pm_TropicalNumber{pm_Min}) = pm_Integer(1)
+orientation(i::Type{<:pm_TropicalNumber{pm_Min}}) = pm_Integer(1)
+orientation(i::pm_TropicalNumber{pm_Max}) = pm_Integer(-1)
+orientation(i::Type{<:pm_TropicalNumber{pm_Max}}) = pm_Integer(-1)
```
kalmarek: add `promote_rule` for `pm_TropicalNumber`s and Integers, Rationals, pm_Rationals

test/tropicalnumber.jl
```diff
@@ -0,0 +1,146 @@
...
+            a = pm_TropicalNumber{A}(5)
+
+            # no copy conversion:
+            @test convert(pm_TropicalNumber{A}, a) === a
+            #
+            # # conversions to Rational types
+            # for T in IntTypes
+            #     RT = Rational{T}
+            #     @test RT(a) isa RT
+            #     @test convert(pm_Rational, RT(a)) isa pm_Rational
```
kalmarek:
```
for R in [Rational{Int}, pm_Rational]
   r = R(1//3)
  @test convert(pm_TropicalNumber{pm_Max}, r) isa pm_TropicalNumber
  tr = convert(pm_TropicalNumber{pm_Max}, r)
  @test convert(R, tr) isa R
  @test convert(R, tr) == r
end
```
 + conversion to Rationals

src/tropicalnumber.jl
```diff
@@ -0,0 +1,70 @@
...
+pm_TropicalNumber{A}() where A <: pm_TropicalNumber_suppAddition = pm_TropicalNumber{A,pm_Rational}()
+
+function Base.:(==)(x::pm_TropicalNumber{A,S},y::Real) where {A <: pm_TropicalNumber_suppAddition, S <: pm_TropicalNumber_suppScalar}
+    return x == pm_TropicalNumber{A,S}(y)
+end
+
+Base.:(==)(x::Real,y::pm_TropicalNumber{A,S}) where {A <: pm_TropicalNumber_suppAddition, S <: pm_TropicalNumber_suppScalar} = (y == x)
+
+#catch wrong typings for basic operations
+function Base.:+(x::pm_TropicalNumber{A1,S1},y::pm_TropicalNumber{A2,S2}) where {A1,A2 <: pm_TropicalNumber_suppAddition, S1,S2 <: pm_TropicalNumber_suppScalar}
```
kalmarek: btw. if we're not promoting anything then this could be shortened to
```julia
Base.promote_rule(::Type{<:pm_TropicalNumber{A, S}}, ::Type{<:pm_TropicalNumber{A, T}}) = pm_TropicalNumber{A, promote_type(S,T)}
for op in (:+, :*, ://, :/, :<)
    @eval begin
        Base.$(op)(x::pm_TropicalNumber{A, S}, y::pm_TropicalNumber{A, T}) where {A,S,T} = $(op)(promote(x,y)...)
        Base.$(op)(x::pm_TropicalNumber{A}, y::pm_TropicalNumber{B}) where {A,B} = 
            throw(DomainError((x,y), "The operation of $(op) for tropical numbers with $A$ and $B$ is not defined"))
    end
end
```
* At the moment the promote rule and the first definition of `Base.$(op)` will never be called as we only allow `pm_Rational` as the second type, but we can cheaply have them for future
* we throw error based on the mismatch of the addition types

src/tropicalnumber.jl
```diff
@@ -0,0 +1,70 @@
...
+const pm_TropicalNumber_suppScalar = Union{pm_Rational}
+
+#convert input to supported Scalar type
+pm_TropicalNumber{A,S}(scalar::Real) where {A <: pm_TropicalNumber_suppAddition, S <: pm_TropicalNumber_suppScalar} = pm_TropicalNumber{A,S}(convert(S,scalar))
+
+#pm_Rational es default Scalar
+pm_TropicalNumber{A}(scalar::Real) where A <: pm_TropicalNumber_suppAddition = pm_TropicalNumber{A,pm_Rational}(scalar)
+pm_TropicalNumber{A}() where A <: pm_TropicalNumber_suppAddition = pm_TropicalNumber{A,pm_Rational}()
+
+function Base.:(==)(x::pm_TropicalNumber{A,S},y::Real) where {A <: pm_TropicalNumber_suppAddition, S <: pm_TropicalNumber_suppScalar}
```
kalmarek: there is no need to specify all of these templates, `where {A, S}` will suffice (nobody will be able to create tropical numbers with different types, as we guard for this in the constructors above. The same in line 15 below

src/tropicalnumber.jl
```diff
@@ -0,0 +1,70 @@
...
+
+Base.:/(x::pm_TropicalNumber{A1,S}, y::pm_TropicalNumber{A2, S}) where{A1, A2, S <: Union{pm_Integer, pm_Rational}} = x//y
+
+function Base.:<(x::pm_TropicalNumber{A1,S1},y::pm_TropicalNumber{A2,S2}) where {A1,A2 <: pm_TropicalNumber_suppAddition, S1,S2 <: pm_TropicalNumber_suppScalar}
+    throw(ArgumentError("comparison of two instances of pm_TropicalNumber only allowed when both parameter types match"))
+end
+
+#zero/one
+
+Base.zero(i::Type{<:pm_TropicalNumber{A}}) where A <: pm_TropicalNumber_suppAddition = zero(pm_TropicalNumber{A}())
```
kalmarek: simple `.. where A = ...` will be sufficient.
Also it's best/nice practice to omit the names of the arguments that are used only for dispatch (`i` is not used on the rhs).

src/tropicalnumber.jl
alexej-jordan: unluckily, this version (and any modification of it i have thought of) gives us an `UndefVarError`:

> ERROR: UndefVarError: op not defined

any idea how to fix this?

kalmarek: forgot about escaping ;-) this should work:
```julia
for op in (:+, :*, ://, :/, :<)
    @eval begin
        Base.$(op)(x::pm_TropicalNumber{A, S}, y::pm_TropicalNumber{A, T}) where {A,S,T} = $(op)(promote(x,y)...)
        Base.$(op)(x::pm_TropicalNumber{A}, y::pm_TropicalNumber{B}) where {A,B} = 
            throw(DomainError((x,y), "The operation of $(esc(op)) for tropical numbers with $(esc(A)) and $(esc(B)) is not defined"))
    end
end
```

alexej-jordan: it still does not work. i should add that the line that is referenced to is not the one you altered but the one starting with `Base.$(op)`

kalmarek: sorry, interpolation is hard ;) Try:
```julia
julia> struct pm_TropicalNumber{A, S} <: Number
          val::S
       end

julia> struct pm_Min end

julia> struct pm_Max end

julia> pm_TropicalNumber{A}(x::S) where {A,S} = pm_TropicalNumber{A, S}(x)

julia> for op in (:+, :*, ://, :/, :<)
           @eval begin
               Base.$(op)(x::pm_TropicalNumber{A, S}, y::pm_TropicalNumber{A, T}) where {A,S,T} = $(op)(promote(x,y)...)
               Base.$(op)(x::pm_TropicalNumber{A}, y::pm_TropicalNumber{B}) where {A,B} = 
                   throw(DomainError((x,y), "The operation $(string($op)) for tropical numbers with $A and $B is not defined"))
           end
       end

julia> t = pm_TropicalNumber{pm_Min}(2//1); s = pm_TropicalNumber{pm_Max}(2.0);

julia> t+s
ERROR: DomainError with (pm_TropicalNumber{pm_Min,Rational{Int64}}(2//1), pm_TropicalNumber{pm_Max,Float64}(2.0)):
The operation + for tropical numbers with pm_Min and pm_Max is not defined
Stacktrace:
 [1] +(::pm_TropicalNumber{pm_Min,Rational{Int64}}, ::pm_TropicalNumber{pm_Max,Float64}) at ./REPL[26]:4
 [2] top-level scope at REPL[28]:1

julia> t*s
ERROR: DomainError with (pm_TropicalNumber{pm_Min,Rational{Int64}}(2//1), pm_TropicalNumber{pm_Max,Float64}(2.0)):
The operation * for tropical numbers with pm_Min and pm_Max is not defined
Stacktrace:
 [1] *(::pm_TropicalNumber{pm_Min,Rational{Int64}}, ::pm_TropicalNumber{pm_Max,Float64}) at ./REPL[26]:4
 [2] top-level scope at REPL[29]:1

```

src/tropicalnumber.jl
```diff
@@ -0,0 +1,70 @@
...
+Base.one(i::Type{<:pm_TropicalNumber{A}}) where A <: pm_TropicalNumber_suppAddition = one(pm_TropicalNumber{A}())
+dual_zero(i::Type{<:pm_TropicalNumber{A}}) where A <: pm_TropicalNumber_suppAddition = dual_zero(pm_TropicalNumber{A}())
+orientation(i::pm_TropicalNumber{pm_Min}) = pm_Integer(1)
+orientation(i::Type{<:pm_TropicalNumber{pm_Min}}) = pm_Integer(1)
+orientation(i::pm_TropicalNumber{pm_Max}) = pm_Integer(-1)
```
saschatimme: Half polymake question: Is this intended to show whether we are in a max setting or in a min setting? If this is the case we should just return `pm_Min` resp. `pm_Max`.

alexej-jordan: but yes, technically seen both approaches are equivalent. but there is a slight difference:
the intention is not to show which setting we are in but to give the user a ready-to-use value corresponding to the setting the tropical number is in. this allows for directly applying maths without having to distinct cases, because real multiplication of the scalar with ´-1´ lets us isomorphically switch between ´pm_Min´ and ´pm_Max´.
this user method also exists within the usual polymake client, but I am not sure whether this is due to the same reasons

src/tropicalnumber.jl
```diff
@@ -0,0 +1,65 @@
...
+for op in (:+, :*, ://, :<)
+    @eval begin
+        Base.$(op)(x::pm_TropicalNumber{A, S}, y::pm_TropicalNumber{A, T}) where {A,S,T} = $(op)(promote(x,y)...)
+        Base.$(op)(x::pm_TropicalNumber{A}, y::pm_TropicalNumber{B}) where {A,B} =
+            throw(DomainError((x,y), "The operation $(string($op)) for tropical numbers with $A and $B is not defined"))
+    end
+end
+
+#at the moment we do not distinct between // and /, so / just refers to //
+Base.:/(x::pm_TropicalNumber{A1,S}, y::pm_TropicalNumber{A2, S}) where{A1, A2, S} = x//y
```
saschatimme: I think this should be
```suggestion
Base.:/(x::pm_TropicalNumber{A}, y::pm_TropicalNumber{A}) where {A} = x//y
```
(you want to enforce same Addition not same element type)

src/tropicalnumber.jl
```diff
@@ -0,0 +1,65 @@
...
+
+#zero/one
+
+Base.zero(::Type{<:pm_TropicalNumber{A}}) where A = zero(pm_TropicalNumber{A}())
+Base.one(::Type{<:pm_TropicalNumber{A}}) where A = one(pm_TropicalNumber{A}())
+dual_zero(::Type{<:pm_TropicalNumber{A}}) where A = dual_zero(pm_TropicalNumber{A}())
+orientation(::pm_TropicalNumber{pm_Min}) = pm_Integer(1)
+orientation(::Type{<:pm_TropicalNumber{pm_Min}}) = pm_Integer(1)
+orientation(::pm_TropicalNumber{pm_Max}) = pm_Integer(-1)
+orientation(::Type{<:pm_TropicalNumber{pm_Max}}) = pm_Integer(-1)
```
saschatimme: I think it is fine if we just return -1 resp. 1 here and not construct an pm_Integer immediately

src/tropicalnumber.jl
```diff
@@ -0,0 +1,65 @@
...
+    if isinf(x)
+        if x > 0
+            return zero(pm_TropicalNumber{pm_Min}())
+        else
+            return dual_zero(pm_TropicalNumber{pm_Min})
+        end
+    else
+        return pm_TropicalNumber{pm_Min}(pm_Rational(x))
+    end
+end
```
saschatimme: Is this correct that we do this to support `pm_TropicalNumber{pm_Min}(Inf)`?

src/tropicalnumber.jl
```diff
@@ -0,0 +1,65 @@
...
+
+Base.:(==)(x::Real,y::pm_TropicalNumber{A,S}) where {A, S} = (y == x)
+
+#catch wrong typings for basic operations
+Base.promote_rule(::Type{<:pm_TropicalNumber{A, S1}}, ::Type{<:pm_TropicalNumber{A, S2}}) where {A, S1, S2} = pm_TropicalNumber{A, promote_type(S1,S2)}
+
+for op in (:+, :*, ://, :<)
+    @eval begin
+        Base.$(op)(x::pm_TropicalNumber{A, S}, y::pm_TropicalNumber{A, T}) where {A,S,T} = $(op)(promote(x,y)...)
```
saschatimme:
```suggestion
        Base.$(op)(x::pm_TropicalNumber{A}, y::pm_TropicalNumber{A}) where {A} = $(op)(promote(x,y)...)
```

src/tropicalnumber.jl
kalmarek: indeed it seems that we do, but `polymake` doesn't complain, so maybe we can leave it as it is?

Looking at the function itself I have two more comments @alexej-jordan 
* this function should return type-stable zeros, without relying on the fact that we allow only `pm_Rational`s as scalars, i.e.
```
function pm_TropicalNumber{pm_Min, S}(x::Float64) where S
    if isinf(x)
        if x > 0
            return zero(pm_TropicalNumber{pm_Min, S}())
        else
            return dual_zero(pm_TropicalNumber{pm_Min, S})
        end
    else
        return pm_TropicalNumber{pm_Min}(S(x))
    end
end
```
* `zero` gets an instance, while `dual_zero` just the type of `pm_TropicalNumber`. why is this so? Even if two calls are equivalent, in library code I'd prefer to see only one of them.

kalmarek:
yep, I agree with everything;

in `polymake` itself `TropicalNumber` is defined in the global namespace (as "number-like" `SmallObject`s), so maybe we keep this as is

src/tropicalnumber.jl
alexej-jordan: you were right, this line had to be updated;
still, what this line is meant to do is calling the respective operations of ´//´ as defined directly above and leave the enforcement of the addition type or the promotion of the scalar type to ´//´. this simplifies our definition of ´/´ to:
´Base.:/(x::pm_TropicalNumber, y::pm_TropicalNumber) = x//y´
if we should decide at some point in the future that we are not satisfied with ´/´ and ´//´ being the same in this setting, we can simply adjust that. the confusion is mainly caused by the different behaviour of the division in the tropical context.

src/tropicalnumber.jl
alexej-jordan: actually, this definition is not used right now. @kalmarek mentioned it would be useful to have this included in the case that we later also allow different scalar types and want to add e.g. a `pm_TropicalNumber{pm_Min,pm_Integer}` to a `pm_TropicalNumber{pm_Min,pm_Rational}`

src/tropicalnumber.jl
alexej-jordan: @saschatimme yes this is correct. polymake itself also allows the construction of a tropicalnumber by passing the string `"inf` or `"-inf"`. as julia has the convenient Float64 values `Inf` and `-Inf`, these seem like a better possibility for our purposes

@kalmarek adjusted it accordingly. the difference you mentioned was most probably a relic of the construction.
also, i made it a bit more compact by distincting between pm_Min and pm_Max via orientation, so we do not have two different functions

src/tropicalnumber.jl
saschatimme: yeah not that important

saschatimme:
Looks good!

PR #186, opened 2020-01-22
Fix/docstrings something something

kalmarek:
* fixes the segfault on asking for docstring (`jlcxx::make_julia_array` issue)
 * return `PolymakeDocstring` which is `print`ed instead of `show`n


src/meta.jl
```diff
@@ -276,7 +276,7 @@ function jl_code(pf::PolymakeFunction)
         end;
         function $(Base.Docs).getdoc(::typeof($(jl_symbol(pf))))
             docstrs = get_docs($func_name, full=true)
-            return Markdown.parse(join(docstrs, "\n\n---\n\n"))
+            return PolymakeDocstring((join(docstrs, "\n\n---\n\n")))
```
benlorenz:
```suggestion
            return PolymakeDocstring(join(docstrs, "\n\n---\n\n"))
```

PR #187, opened 2020-01-23
*WIP* trying to remove pm_ and update to polymake 4.0

benlorenz:
removed all pm_ prefixes for the types, adapted to pm::Int, removed a few int32 types, adapt to BigObject rename.
this is still somewhat broken due to conflicts with Base types and needs a lot of cleanup.
do we really want to remove all pm_ prefixes?

Still needs lots of cleanup...

PS: this needs a source installation of polymake 4.0 / master


benlorenz:
obsolete

PR #190, opened 2020-01-23
Remove void calls

saschatimme:
This was introduced for getting visualization working, but by now this seems to be never called....
So let's get rid of it.


saschatimme:
Also sneak in a small fix of an (unused) perl rule

kalmarek:
`+7 -30` I like this!

Issue #196, opened 2020-01-27
Sparse system tests flaky


saschatimme:
This test 

https://github.com/oscar-system/Polymake.jl/blob/ca21f7e85f772b9fecd7ac3aacefb44ebc93cfd0/test/sparsematrix.jl#L392-L400

is quite flaky on CI which seems ... odd.

kalmarek:
it might have to do with floating point zeros (sme subnormals are treated as zeros?)
@alexej-jordan 
in the worst case we can add `Random.seed!(1)`, but that only masks the issue which should be better solved

saschatimme:
yeah let's put a seed at the beginning front for now...

Also why do we have such a big test matrix?

kalmarek:
I'd prefer a real solution though; The size is @alexej-jordan s invention; it allocates a quarter of million entries, which seems a little bit excessive; especially that my next aim is at speeding-up tests

saschatimme:
Sure, I will leave #197 in case we want to have a (temporarily) quick fix.
@alexej-jordan a matrix even with only 1% non-zero entries is quite often not considered sparse enough to justify sparse matrix storage format

alexej-jordan:
I'm on it, will now try to reproduce. do we have any additional information from CI why this test fails? @saschatimme 
additionally to the idea @kalmarek mentioned i could also think of some problem when sorting the entries, but i thought this should work out due to the way the entries are sorted within their corresponding sparse matrix structure.
actually, the big size of the test matrix is a relic of trying to figure this out; i wanted a good chance that some rows and/or columns contain several entries

alexej-jordan:
after some more testing, @kalmarek seems to have called the problem by its name: subnormal numbers.
```
julia> eps(Float32)
1.1920929f-7
```

and every time i could reproduce the failure it was caused by the only entry in `jsm` which was below the mentioned epsilon, resulting in the corresponding entry of `psm` being zero.
setting the same value within a dense `Polymake.Matrix{Float64}` works perfectly fine which leads me to assume that the internal `iszero` method of polymake might use this epsilon, or maybe somewhere in the process of setting the entry.
of course, we could quickly make the tests work by forcing Float32 values for the random sparse matrix, but it would be better if we fully support Float64 in `Polymake.SparseMatrix` the same as in `Polymake.Matrix`

benlorenz:
The exact value is slightly different `1.0e-7`, see `lib/core/src/inlines.cc`.

[...]

PR #197, opened 2020-01-27
Fix random seed in SparseArray test to fix flaky ci

saschatimme:
Temporary fix for #196


kalmarek:
lets merge this only if finding the real issue turns out to be too time-consuming. @alexej-jordan could you have a look at this?

PR #200, opened 2020-01-28
Fix sparsematrix test

### LINKED ISSUE #196: Sparse system tests flaky

alexej-jordan:
by making `pm::local_epilson_keeper` accessible from julia we can adjust the precision of polymake. this fixes the flaky sparse matrix test where subnormal numbers were identified as zero.

to have this epsilon work, the instance with the desired value must exist. thats why i define the global variable `_pm_local_epsilon_keeper` in the `__init__()` method

This closes #196

### DIFF
--- a/deps/src/polymake_sparsematrix.cpp
+++ b/deps/src/polymake_sparsematrix.cpp
@@ -8,6 +8,11 @@
 
 void polymake_module_add_sparsematrix(jlcxx::Module& polymake)
 {
+
+    polymake
+        .add_type<pm::local_epsilon_keeper>("local_epsilon_keeper")
+        .constructor<double>();
+
     polymake
         .add_type<jlcxx::Parametric<jlcxx::TypeVar<1>>, jlcxx::ParameterList<jlcxx::TypeVar<1>,int>>(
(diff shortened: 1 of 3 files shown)


saschatimme:
These are not subnormal numbers. A subnormal Float32 is smaller than 2^-126 = 1.1e-38 and a subnormal Float64 number is smaller than 2^-1022 = 2.2e-308. This now just decreases the likelihood that this happens.

I would suggest two options to fix this:

1) Just have a non-deterministic test where all entries are larger than 1.

2) Make it possible to read the global epsilon from polymake and use [SparseArrays.droptol!](https://docs.julialang.org/en/v1/stdlib/SparseArrays/#SparseArrays.droptol!) with that value in the test.

kalmarek:
I'd prefer to be able to query `global_epsilon_keeper` (it's global right ?:), and `droptol!` on the test matrix (what Sascha proposed in 2)

PR #205, opened 2020-01-30
Adjust Travis Test Matrix

saschatimme:
~~Everything on 1.3 should also work on 1.2. 1.4 has already a first release candidate, so we can use this on CI already~~
**Update:** Add 1.4 to the test matrix. Also exclude all mac jobs but for 1.3. Travis ci mac tests are really slow compared to linux (not much parallelism and slow scheduling).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,20 @@ julia:
   - 1.0
   - 1.2
   - 1.3
+  - 1.4
   - nightly
 matrix:
- allow_failures:
- - julia: nightly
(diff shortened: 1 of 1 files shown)


benlorenz:
We might want to keep 1.2 in there because of Oscar and Singular.jl which so far don't support 1.3.

saschatimme:
But then let me exclude all but one of the Mac test (for 1.3) since those are particular slow on Travis.

PR #213, opened 2020-02-07
Macro/convert to

alexej-jordan:
@saschatimme the macro is working for proper input and also detects wrong input, but i could not find a way to include the latter in the error message


saschatimme:
```julia
macro convert_to(args...)
    # Catch case that only one or more than two arguments are given
    if length(args) != 2
        throw(ArgumentError("@convert_to needs to be called with 2 arguments, e.g., `@convert_to Matrix{Integer} A`."))
    end
    expr1, expr2 = args
    :(
        try
            # expr2 needs to be escaped
            @pm common.convert_to{$expr1}($(esc(expr2)))
        catch ex
            # To not catch things like UndefVarError only catch ErrorException
            # since this is currently thrown if something invalid is parsed.
            if ex == ErrorException
                # Use QuoteNodes to keep expr1 and expr2 as Expr around
                expr1 = $(QuoteNode(expr1))
                expr2 = $(QuoteNode(expr2))
                throw(ArgumentError("Can not parse the expression passed to @convert_to macro:\n$expr1 $expr2\n Only `@convert_to PerlType argument` syntax is recognized"))
            else
                rethrow(ex)
            end
        end
    )
end
```

saschatimme:
Also a doc string would be good.

alexej-jordan:
@saschatimme even though your code is very much onto the point i have a question and an alteration (in combination with a question):
1) how do i force an `ErrorException`?
2) to properly catch the `ArgumentError` i also had to adjust the way it is thrown; that is now done within an expression. otherwise the macro would even though continue to do its work and fail. thats why i had to move the main code into the `else` block. is there any cleaner way to accomplish this?

src/convert.jl
```diff
@@ -70,3 +71,57 @@ function promote_to_pm_type(::Type{<:Union{Vector, Matrix, SparseMatrix}}, S::Ty
...
+    else
+        expr1, expr2 = args
+        :(
+            try
+                # expr2 needs to be escaped
+                @pm common.convert_to{$expr1}($(esc(expr2)))
+            catch ex
+                # To not catch things like UndefVarError only catch ErrorException
+                # since this is currently thrown if something invalid is parsed.
+                if ex == ErrorException
```
saschatimme: Sorry I had this wrong in my snippet. This should be that:
```suggestion
                if ex isa ErrorException
```

saschatimme:
> @saschatimme even though your code is very much onto the point i have a question and an alteration (in combination with a question):
> 
> 1. how do i force an `ErrorException`?

```julia
error("my_err")
```
> 2. to properly catch the `ArgumentError` i also had to adjust the way it is thrown; that is now done within an expression. otherwise the macro would even though continue to do its work and fail. thats why i had to move the main code into the `else` block. is there any cleaner way to accomplish this?

Really?
```julia
julia> macro convert_to(args...)
           # Catch case that only one or more than two arguments are given
           if length(args) != 2
               throw(ArgumentError("@convert_to needs to be called with 2 arguments, e.g., `@convert_to Matrix{Integer} A`."))
           end
           expr1, expr2 = args
           :(
               try
                   # expr2 needs to be escaped
                   @pm common.convert_to{$expr1}($(esc(expr2)))
               catch ex
                   # To not catch things like UndefVarError only catch ErrorException
                   # since this is currently thrown if something invalid is parsed.
                   if ex == ErrorException
                       # Use QuoteNodes to keep expr1 and expr2 as Expr around
                       expr1 = $(QuoteNode(expr1))
                       expr2 = $(QuoteNode(expr2))
                       throw(ArgumentError("Can not parse the expression passed to @convert_to macro:\n$expr1 $expr2\n Only `@convert_to PerlType argument` syntax is recognized"))
                   else
                       rethrow(ex)
                   end
               end
           )
       end
@convert_to (macro with 1 method)

julia> @convert_to a b 3
ERROR: LoadError: ArgumentError: @convert_to needs to be called with 2 arguments, e.g., `@convert_to Matrix{Integer} A`.
Stacktrace:
 [1] @convert_to(::LineNumberNode, ::Module, ::Vararg{Any,N} where N) at ./REPL[11]:4
in expression starting at REPL[12]:1

julia> @convert_to a
ERROR: LoadError: ArgumentError: @convert_to needs to be called with 2 arguments, e.g., `@convert_to Matrix{Integer} A`.
Stacktrace:
 [1] @convert_to(::LineNumberNode, ::Module, ::Vararg{Any,N} where N) at ./REPL[11]:4
in expression starting at REPL[14]:1
```

alexej-jordan:
yes, i even just re-tried the tests on your version just to be sure. the difference is very subtle: seemingly, the `ArgumentError` is nested within a `LoadError`, most probably because the first causes the latter when trying to parse the expression.
you can have a look at the tests, but i think they are okay.
still, one new problem arose in the meantime:
i merged the new master with the updated cxx support into this branch and now the conversion `@convert_to Array{Set{Int}} [Set([1, 2, 4, 5, 7, 8]), Set([1]), Set([6, 9])]` does not work anymore:
```julia
julia> @convert_to Array{Set{Int}} [Set([1, 2, 4, 5, 7, 8]), Set([1]), Set([6, 9])]
ERROR: Cannot pass function value: conversion failed for argument of type ArrayAllocated
Stacktrace:
 [1] internal_call_function(::String, ::CxxWrap.StdLib.StdVectorAllocated{CxxWrap.StdLib.StdString}, ::Array{Any,1}) at /home/jordan/.julia/packages/CxxWrap/GxGUV/src/CxxWrap.jl:552
 [2] #call_function#17(::Array{String,1}, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(Polymake.call_function), ::Type{PropertyValue}, ::Symbol, ::Symbol, ::Array{Set{Int64},1}) at /home/jordan/.julia/dev/Polymake/src/call_function.jl:21
 [3] (::Polymake.var"#kw##call_function")(::NamedTuple{(:template_parameters,),Tuple{Array{String,1}}}, ::typeof(Polymake.call_function), ::Type{PropertyValue}, ::Symbol, ::Symbol, ::Array{Set{Int64},1}) at ./none:0
 [4] #call_function#18 at /home/jordan/.julia/dev/Polymake/src/call_function.jl:27 [inlined]
 [5] (::Polymake.var"#kw##call_function")(::NamedTuple{(:template_parameters,),Tuple{Array{String,1}}}, ::typeof(Polymake.call_function), ::Symbol, ::Symbol, ::Array{Set{Int64},1}) at ./none:0
 [6] top-level scope at /home/jordan/.julia/dev/Polymake/src/convert.jl:141
```

is there anything im missing? i would not have thought that the `pm` and the `convert_to` macro rely on the wrapped typing in this way

kalmarek:
unfortunately this has nothing to do with the macro, and everything to do with `CxxLong` stuff:
in `/deps/src/generated/polymake_call_function_feed_argument.h` line 66 reads
```c
else if (jl_subtype(current_type, POLYMAKETYPE_Array_Set_Int)) {
   ...
}
```
The problem is that now the argument becomes `Polymake.ArrayAllocated{Polymake.Set{CxxWrap.CxxLong}}}` which is not a subtype of `Array_Set_Int` anymore which is `Polymake.Array{Polymake.Set{Int}}`...

kalmarek:
@alexej-jordan to fix `convert_to` function please
1. `%s/{CxxWrap.CxxLong}/{Int64}/g` on `deps/type_setup.jl`
2. `Pkg.build("Polymake")`
3. check that
```julia
julia> Polymake.call_function(:common, :convert_to, [Set([1, 2, 4, 5, 7, 8]), Set([1]), Set([6, 9])]; template_parameters=["Array<Set<Int>>"])
pm::Array<pm::Set<long, pm::operations::cmp>>
{1 2 4 5 7 8}
{1}
{6 9}
```

kalmarek:
@alexej-jordan possible tests:

```julia
julia> @test_throws LoadError eval(:(@convert_to a b 3))
Test Passed
      Thrown: LoadError

julia> err = try 
                eval(:(@convert_to a b 3))
              catch err
                err
              end;

julia> @test err.error isa ArgumentError
Test Passed

```

saschatimme:
I added some finishing touches and this should now be good to merge.

test/convert.jl
```diff
@@ -25,4 +25,30 @@
 
         @test polytope.cube(MyInt(3)) isa Polymake.BigObject
     end
+
+    @testset "@convert_to" begin
+        @convert_to Integer 64 isa Polymake.Integer
```
benlorenz:
```suggestion
        @test (@convert_to Integer 64) isa Polymake.Integer
```

README.md
```diff
@@ -285,7 +285,7 @@ and so should be accepted by all methods that apply to the abstract types.
 yet, you will not be able to access its content or in general use it **from Julia**,
 however you can always pass it back as an argument to a `polymake` function.
 Moreover you may try to convert to Julia understandable type via macro
-`@pm common.convert_to{C++{templated, type}}(obj)`.
+`@convert_to C++{template, names} obj`.
```
benlorenz:
```suggestion
`@convert_to SomeType{Template, Names} obj`.
```

PR #231, opened 2020-03-09
remove (empty) codecov token from github workflow

benlorenz:
should not be needed anymore


benlorenz:
removing the token doesn't seem to help for the http error 400 during upload so we keep this open for now.

PR #234, opened 2020-03-10
Add support for Array<Polynomial>

saschatimme:
This was more complicated than expected since adding support for this
initially created a wonderful dependency cycle.

I resolved this by splitting the array declarations into two parts.
This introduces some code duplication but I don't see how this is
avoidable without pulling out the big guns...


saschatimme:
Travis failure on 1.4 is again the CxxWrap segfault...

PR #238, opened 2020-03-17
Drop Julia < 1.3 + Upgrade to CxxWrap 0.10

saschatimme:
See discussion in https://github.com/oscar-system/Oscar.jl/issues/80

**Update:** I also upgraded to CxxWrap 0.10


benlorenz:
Hm, most of the crashes for 1.4 are the known CxxWrap precompilation error but I dont like this one here https://github.com/oscar-system/Polymake.jl/runs/513326853?check_suite_focus=true
First there is a not really relevant `msg` missing in BoundsError but the main error is somewhere in the string conversion from `shell_execute` via `StdString`:
```
caused by [exception 1]
BoundsError: attempt to access CxxWrap.StdLib.StdStringAllocated
  at index [29]
Stacktrace:
 [1] getproperty(::CxxWrap.StdLib.StdStringAllocated, ::Symbol) at ./Base.jl:33
 [2] _make_ref(::Type{CxxWrap.CxxWrapCore.ConstCxxRef}, ::CxxWrap.StdLib.StdStringAllocated) at /home/runner/.julia/packages/CxxWrap/lDNAy/src/CxxWrap.jl:225
 [3] CxxWrap.CxxWrapCore.ConstCxxRef(::CxxWrap.StdLib.StdStringAllocated) at /home/runner/.julia/packages/CxxWrap/lDNAy/src/CxxWrap.jl:230
 [4] unsafe_convert(::Type{CxxWrap.CxxWrapCore.ConstCxxRef{CxxWrap.StdLib.StdString}}, ::CxxWrap.StdLib.StdStringAllocated) at /home/runner/.julia/packages/CxxWrap/lDNAy/src/StdLib.jl:107
 [5] c_str(::CxxWrap.StdLib.StdStringAllocated) at /home/runner/.julia/packages/CxxWrap/lDNAy/src/CxxWrap.jl:552
 [6] String(::CxxWrap.StdLib.StdStringAllocated) at /home/runner/.julia/packages/CxxWrap/lDNAy/src/StdLib.jl:94
 [7] convert(::Type{String}, ::CxxWrap.StdLib.StdStringAllocated) at ./strings/basic.jl:210
 [8] convert(::Type{Tuple{String,String,String}}, ::Tuple{CxxWrap.StdLib.StdStringAllocated,CxxWrap.StdLib.StdStringAllocated,CxxWrap.StdLib.StdStringAllocated}) at ./essentials.jl:310 (repeats 2 times)
 [9] shell_execute(::String) at /home/runner/work/Polymake.jl/Polymake.jl/src/util.jl:19
 [10] __init__() at /home/runner/work/Polymake.jl/Polymake.jl/src/Polymake.jl:63
```

Also it seems that I need to remove some required checks from the github configuration.

.github/workflows/runtests.yml
```diff
@@ -7,10 +7,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['1.0', '1.3']
+        julia-version: ['1.3']
```
kalmarek:
```suggestion
        julia-version: ['1.3', '1.4']
```

.github/workflows/runtests.yml
```diff
@@ -22,15 +22,15 @@ jobs:
...
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-  
+
   test-julia-nightly:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['^1.4.0-rc1', 'nightly']
+        julia-version: ['^1.4.0-rc2', 'nightly']
```
kalmarek:
```suggestion
        julia-version: ['nightly']
```

saschatimme:
CxxWrap master seems to work with 1.4 reliably and a new release is also probably happening soon https://github.com/JuliaInterop/CxxWrap.jl/issues/214#issuecomment-602278389

kalmarek:
after three restarts, all tests seem to pass ;-) but it seems to me that this should be `v0.4.0` ?

benlorenz:
I think we should wait for the cxxwrap release before merging this, I hope it gets better then.

saschatimme:
Yeah definitely let's wait until CxxWrap is released.

And sure we can also make this 0.4 if we want...

fingolfin:
@saschatimme @benlorenz the CxxWrap release in turn seems to wait for somebody (you?) to confirm that the changes in its master branch actually help. See https://github.com/JuliaInterop/CxxWrap.jl/issues/214 and https://github.com/JuliaInterop/CxxWrap.jl/issues/223

kalmarek:
@fingolfin I will try to do this this week, but I have way too much stuff on my head right now

fingolfin:
So it seems they released CxxWrap 0.10.0... I guess we'll want to switch to that then here and in Singular.jl ...

.github/workflows/runtests.yml
```diff
@@ -1,13 +1,19 @@
...
+     branches:
+       - master
 
 jobs:
   test-julia-stable:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['1.3', '1.4']
+        julia-version: ['1.4', 'nightly']
```
kalmarek, line 16: what about `1.3`, we aren't dropping it as well?

saschatimme: No, but I thought 1.4 would be sufficient, but I can add it back

saschatimme:
The CxxWrap update is unfortunately not just a version bump, we now get

```julia
ERROR: LoadError: UndefVarError: jlcxx_path not defined
```

during the build

benlorenz:
> The CxxWrap update is unfortunately not just a version bump, we now get
> 
> ```julia
> ERROR: LoadError: UndefVarError: jlcxx_path not defined
> ```
> 
> during the build.

See https://github.com/JuliaInterop/CxxWrap.jl/issues/225, it is already fixed in master but we need to use `dirname(dirname(CxxWrap.CxxWrapCore.libcxxwrap_julia_jll.libcxxwrap_julia_path))` for now.

saschatimme:
> > The CxxWrap update is unfortunately not just a version bump, we now get
> > ```julia
> > ERROR: LoadError: UndefVarError: jlcxx_path not defined
> > ```
> > 
> > 
> > during the build.
> 
> See [JuliaInterop/CxxWrap.jl#225](https://github.com/JuliaInterop/CxxWrap.jl/issues/225), it is already fixed in master but we need to use `dirname(dirname(CxxWrap.CxxWrapCore.libcxxwrap_julia_jll.libcxxwrap_julia_path))` for now.

Just tried this locally, but CMake isn't happy with that :/

fingolfin:
Please take a look at https://github.com/oscar-system/Singular.jl/pull/216 to see how I did it for Singular.jl -- that patch actually makes the build work with both CxxWrap 0.9 and 0.10. It is based on the sample code CxxWrap provides.

saschatimme:
@kalmarek  @benlorenz  Can maybe one of you pick this up?

benlorenz:
> @kalmarek @benlorenz Can maybe one of you pick this up?

yes

kalmarek:
@benlorenz this works for me:
```julia
jlcxx_cmake_dir = joinpath(
    dirname(CxxWrap.CxxWrapCore.libcxxwrap_julia_jll.libcxxwrap_julia_path),
    "cmake", 
    "JlCxx")
```
? Also references to StaticlyTypedNumbers (in `polymake_rationals`) need to be changed to values

benlorenz:
> @benlorenz this works for me:
> 
> ```julia
> jlcxx_cmake_dir = joinpath(
>     dirname(CxxWrap.CxxWrapCore.libcxxwrap_julia_jll.libcxxwrap_julia_path),
>     "cmake", 
>     "JlCxx")
> ```
> 
> ? Also references to StaticlyTypedNumbers (in `polymake_rationals`) need to be changed to values

I had one extra `dirname` in there but an additional `lib`, StrictlyTypedNumbers are adjusted now. First testrun failed because of some http download timeout error, restarted now.

fingolfin:
Could there perhaps be first a release of the current master branch, before this PR is merged, as that fixes an issue people were having when installing Polymake.jl (due to the "too new" libstdc++)?

fingolfin:
I mean I can call the registrator myself of course, but I don't want to mess with your package behind your back :-)

kalmarek:
ok, so now the question is should we keep the compatibility with CxxWrap-0.9, or drop it (there are a few workarounds in the code for bugs in 0.9).

If we keep the compatibility this looks good to me;

kalmarek:
@fingolfin good point; If there no objections, I'll tag on https://github.com/oscar-system/Polymake.jl/commit/fd04c00ecc27a282666acb07aeed32227d0b9914

benlorenz:
The new `jlcxx_cmake_path` is not compatible with 0.9 so we would need to add some version-dependent code there. And we should probably add a test for 0.9 on github actions then?

benlorenz:
> @fingolfin good point; If there no objections, I'll tag on [fd04c00](https://github.com/oscar-system/Polymake.jl/commit/fd04c00ecc27a282666acb07aeed32227d0b9914)

fine with me

Project.toml
```diff
@@ -18,9 +18,9 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 [compat]
 BinaryProvider = "^0.5.3"
 CMake = "^1.1.1"
-CxxWrap = "^0.9.0"
+CxxWrap = "0.9,0.10"
```
fingolfin, line 21: Is this really meant to be compatible with both? If so, why is the version of the package even changed from 0.3.3 to 0.4.0, and not just to 0.3.4?

Project.toml
```diff
@@ -18,9 +18,9 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
...
 JSON = "^0.20, ^0.21"
-julia = "^1.0"
+julia = "^1.3"
```
fingolfin, line 23: Just out of curiosity: since `^` is the default anyway, is there a reason you specify it explicitly in most of these compat entries?

benlorenz:
> ok, so now the question is should we keep the compatibility with CxxWrap-0.9, or drop it (there are a few workarounds in the code for bugs in 0.9).
> 
> If we keep the compatibility this looks good to me;

There is probably no point in supporting 0.9.0 unless we want to keep compatibility with julia 1.0?

benlorenz:
lets see if 0.9 still works
and test xcode 11.3.1 and 11.4 (the latter should fail)

benlorenz:
Tests are more or less as expected: XCode 11.4 fails and nightly has some issues:
```
  MethodError: no method matching mul_with_overflow(::Polymake.IntegerAllocated, ::Polymake.IntegerAllocated)
```
But otherwise cxxwrap 0.9 and 0.10 work fine.

.github/workflows/runtests.yml
```diff
@@ -1,34 +1,37 @@
...
       matrix:
-        julia-version: ['1.0', '1.3', '1.4']
+        julia-version: ['1.3', '1.4', 'nightly']
         os: [ubuntu-latest, macOS-latest]
-      fail-fast: true
+        xcode: [ '11.4', '11.3.1' ]
+        cxxwrap: [ '0.9', '0.10' ]
+        exclude:
+        # only julia 1.4 on macOS
+        # choose one arbitrary xcode version for linux
```
fingolfin, line 22: huh? xcode is macos specific, what does "choosing an xcode version for linux" even mean?

fingolfin:
This PR seems to do (at least) three independent things:
- drop support for Julia < 1.3 -- a tiny patch that could be merged immediately
- drop use of Travis CI and some changes to GitHub actions -- probably could also be merged immediately
- adds support for CxxWrap 0.10 and some related work -- maybe needs a little bit more time to brew?

Perhaps it would make sense to split that up?

.github/workflows/runtests.yml
benlorenz: the environment variable will be ignored but I cannot expand just one entry of an axis in a new dimension for the xcode version (at least I dont know of a quick way to do this in github yaml)

benlorenz:
> This PR seems to do (at least) three independent things:
> 
> * drop support for Julia < 1.3 -- a tiny patch that could be merged immediately

I included the version bump to 0.4.0 with this pull as it breaks compatibility.

> * drop use of Travis CI and some changes to GitHub actions -- probably could also be merged immediately

Doesn't really make sense until the cxxwrap 0.10 changes are merged. The xcode 11.4 change is more an extra to keep this on the radar and really not urgent.

> * adds support for CxxWrap 0.10 and some related work -- maybe needs a little bit more time to brew?

I dont think cxxwrap 0.10.0 needs extra time.

> Perhaps it would make sense to split that up?

I split it into three.


PS: Merging some of these might require admin permissions as we changed the test-names which are listed in the branch protection rules (or tests retriggered afterward changing the rules)

PR #240, opened 2020-03-23
deps: switch from Nemo to LoadFlint

benlorenz:
to avoid nemo dependency and avoid duplicate flint and gmp libraries.

cc: @fingolfin @fieker


Project.toml
```diff
@@ -20,7 +20,6 @@ BinaryProvider = "^0.5.3"
 CMake = "^1.1.1"
 CxxWrap = "^0.9.0"
 JSON = "^0.20, ^0.21"
-Nemo = "^0.16"
```
kalmarek: LoadFlint = "^0.1"

benlorenz: done

PR #241, opened 2020-03-25
supress banner in oscar

fieker:


kalmarek:
@fieker please convince also @wbhart to add an option to suppress the banner :D

fieker:
On Wed, Mar 25, 2020 at 03:02:51AM -0700, kalmarek wrote:
> @fieker please convince also @wbhart to add an option to suppress the banner :D

Long since included. When loading Nemo in Oscar (or Hecke) no banner..
Same technology as for polymake.
Similar Hecke prints no banner in Oscar either

But Oscar has a beautiful banner (or wil have once we get it done)

Bill's problem is copyright notices, but we promise to have them
(printed/ accessible) in Oscar, thus don't need them individually

This is not a blanked ban, nor a general mechanism. Its just that within
our projects, where we will print apropriate banners, we suppress them

kalmarek:
indeed! https://github.com/Nemocas/Nemo.jl/commit/4ecd9a92c65b2b4b92bc67a5556dd011ca333176

PR #243, opened 2020-03-31
?Polymake

fieker:
add a doc string for the Polymake module, allows, eg. in Oscar to do
?Polymake


benlorenz:
obsolete with #249

PR #245, opened 2020-04-01
Documentation/new

alexej-jordan:
with the help of `Documenter.jl` the documentation can be generated. i mostly sorted the information from our big readme file into some pages.

still, when it comes to the tables, HTML within this markdown does not work properly (`<br>` is not translated), but using the `@raw html` macro from `Documenter.jl`  has some problems with the $-sign. Using the HTML version `&#36;` does not work either.


kalmarek:
yeah, tables are always a pain;
how about using `<code>$p</code>` instead of `$p` inside html?

alexej-jordan:
the `<code>` blocks worked so i put in the effort to convert the tables to html. i am not really satisfied with the looks of the table but having a hard time accessing the attributes via css in the `@raw html` blocks

kalmarek:
this looks already ok;
one way of making it better (sorry for this, I'm this kind of guy:D) is to use
`jldoctest` instead of code blocks https://juliadocs.github.io/Documenter.jl/stable/man/doctests/
this way we always ensure that the examples in documentation are in sync with the library.

alexej-jordan:
this pull request is now expanded by:
* docstrings for polydb module
* section about the polydb within the documentation
* workflow for creating the documentation (incomplete)

@benlorenz you probably know more about how to integrate this the most efficient and stable way. from my understanding, we keep the automated documentation on a different branch allowing us to run this workflow on every push of that branch when we intend to. is this the correct way? and how do we make the generated files available?

benlorenz:
I don't have experience with this setup but I think we should just follow: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/
So you need to generate that ssh key and adjust the github action. And I can then add it to the repository and adjust the settings for the gh-pages branch.

benlorenz:
I enabled pull request preview builds and the result can be found here: https://oscar-system.github.io/Polymake.jl/previews/PR245/

docs/src/index.md
```diff
@@ -0,0 +1,49 @@
+# Polymake.jl
+
+`Polymake.jl` is a Julia package for using [`polymake`](https://polymake.org/doku.php), a software for research in polyhedral geometry from Julia.
```
saschatimme, line 3:
```suggestion
[`Polymake.jl`](https://github.com/oscar-system/Polymake.jl) is a Julia package for using [`polymake`](https://polymake.org/doku.php), a software for research in polyhedral geometry from Julia.
```

Always nice to have a backlink to the repo :)

alexej-jordan:
seems like this workflow does what it is supposed to do. the `deploydocs` method of `Documenter.jl` does a pretty good job.

some checks did not pass for the recent commit due to some error with julia, but this appears to have some external cause.

are there any changes that need to be applied before we can merge this? (besides the suggestion by @saschatimme and besides changing the branch in the workflow to `master`)

oh, and thank you @benlorenz by the way, for always having an eye on this and for preparing the `gh-pages` branch. when this pull request is merged, you just need to adjust the settings on github (https://docs.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site), as i do not have the permission to access these settings.

benlorenz:
The github settings should be done already.

You could remove most of the readme.md now, just keep the first section with updated links (will be broken until the pull request is done) and finally you need to mark the pull request ready for review.

These errors on 1.3 are unrelated but really weird, the version numbers and hashes for all dependencies are the same as in another successful build 11 hours ago.

benlorenz:
Unless there are any objections, I will merge this later today.

PR #252, opened 2020-04-10
fix #251: adjust ostringstream constructor for recent libstdc++

### LINKED ISSUE #251: Polymake.jl doesn't work if user has too new C++ library
The following was reported by Meinolf Geck on the oscar mailing list, but I've seen it elsewhere before, too: when using Julia 1.3.1 under Ubuntu 19.10, install Oscar seems to work, but loading it doesn't. This can be reduced to Polymake.jl not loading:
```
julia> using Oscar
[ Info: Precompiling Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13]
ERROR: LoadError: could not load library "/home/geck/.julia/packages/Polymake/kIMyn/src/../deps/src/libpolymake.so"
/usr/local/lib/julia-1.3.1/bin/../lib/julia/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /home/geck/.julia/packages/Polymake/kIMyn/src/../deps/src/libpolymake.so)
Stacktrace:
...
```

Here's my understanding of the issue: Julia includes a copy of a the standard C++ library (which is fine; and it has to, as the user might not have it, or might have a version that's too old). On your computer, there is a *newer* version of that library; now when Polymake.jl is installed (which involves CxxWrap.jl, and compiling C++ code), the compiler generates output which assumes this newer version of the C++ library will be loaded. But in reality, when run under Julia, it always loads the older version of the C++ library shipped with Julia. As a result, the freshly compiled code fails to load.

So, how can we fix this? One view would be that `Polymake.jl`'s build system should explicitly use the same C++ lib as Julia does. Perhaps also CxxWrap needs a corresponding change (I am not sure whether other packages using CxxWrap are affected?).

Telling users to remove the Julia copy of `libstdc++` IMHO is not a viable solution. But perhaps you have some other ideas or already a fix? 

The issue also exists in Julia 1.4 and on other Linux distributions; it's really just a matter of how old the C++ library is.

benlorenz:
Better fix this time, just 4 characters. As it turns out, only the default constructor for ostringstream requires `GLIBCXX_3.4.26`, we can just use an empty string instead.

---- obsolete first try:
@fingolfin @kalmarek 

Unfortunately just dlopening a new `libstdc++.so.6.0.26` doesn't help because the `DT_NEEDED` of `libpolymake.so` is just `libstdc++.so.6` which will still resolve to the old one from julia which is also in memory.

But since we love having more and more dependencies we can patch the library with `patchelf` to use the full filename of the correct `libstdc++`. I tried looking for options to gcc to do something similar at link time but that seems impossible.

PS: all of the above applies only to Sys.islinux()

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Polymake"
 uuid = "d720cf60-89b5-51f5-aff5-213f193123e7"
 repo = "https://github.com/oscar-system/Polymake.jl.git"
-version = "0.3.2"
+version = "0.3.3"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
--- a/deps/src/polymake_functions.h
+++ b/deps/src/polymake_functions.h
@@ -20,7 +20,7 @@ pm::Integer new_integer_from_bigint(jl_value_t*);
 template <typename T>
 std::string show_small_object(const T& obj, bool print_typename = true)
 {
-    std::ostringstream buffer;
+    std::ostringstream buffer("");
     auto               wrapped_buffer = wrap(buffer);
     if (print_typename) {
         wrapped_buffer << polymake::legible_typename(typeid(obj)) << pm::endl;


benlorenz:
Unfortunately adding this new libstdc++ broke exactly the show method mentioned in #251
```
Polymake.Vector{Int64}: Test Failed at /home/runner/work/Polymake.jl/Polymake.jl/test/vectors.jl:88
  Expression: string(V) == "pm::Vector<long>\n5 10 3"
   Evaluated: "pm::Vector<long>\n" == "pm::Vector<long>\n5 10 3"
```

kalmarek:
@benlorenz I tried this and see no adverse effects:
* `strings libpolymake.so | fgrep GLIBCXX_3.4.26 | c++filt` is empty
* `Polymake.Vector([1,2,3])` prints correctly (and all the tests pass).

benlorenz:
The print bug was with my previous try where I pulled in the new libstdc++.

I added a version bump so we can directly make a new release if needed, feel free to merge this as soon as the tests are done.

PR #253, opened 2020-04-14
Travis: simplify config, add Xcode 11.4 test

fingolfin:
The test with Xcode 11.4 of course fails (see https://github.com/oscar-system/Oscar.jl/issues/82)


fingolfin:
Made redundant by #256

PR #254, opened 2020-04-14
CxxWrap 0.10.0

benlorenz:
part of #238


Project.toml
```diff
@@ -18,7 +18,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 [compat]
 BinaryProvider = "^0.5.3"
 CMake = "^1.1.1"
-CxxWrap = "^0.9.0"
+CxxWrap = "0.9,0.10"
```
fingolfin, line 21: So a Polymake.jl compiled against CxxWrap 0.9 can be runtime linked against 0.10 and works, and the other way around? But then why did they make a semver change?

benlorenz: Nope, building against 0.9 and using it with 0.10 doesn't work (just tested and fails with some missing symbol), but there is no way to automatically enforce a Pkg.build anyway.
0.9.x is supposed to be the LTS branch.

fingolfin: Well, then I don't understand how the above compat line is valid? We need to pick one, don't we?

Independently of this: does 0.10.0 actually give us any benefits? I.e. are there any tests (e.g. with nightly) that previously failed and now pass? I still see failing tests reported below.

Finally, where can one learn that 0.9 is the LTS? I couldn't find this info in the CxxWrap repo. If it is an LTS, I hope they'll make a 0.9.1 released with a fixed `prefix_path()` (and also 0.10.1 with its own fix)

kalmarek: The line is valid, Polymake.jl can build and work against both versions of CxxWrap, that's the only thing the line says, I believe

saschatimme: > Independently of this: does 0.10.0 actually give us any benefits? I.e. are there any tests (e.g. with nightly) that previously failed and now pass? I still see failing tests reported below.

We see repeated CI failures on 1.4 with CxxWrap 0.9, these seem to be gone now. The nightly is a new error unrelated to CxxWrap.

kalmarek: btw. nightly failure is caused by this one
https://github.com/JuliaLang/julia/commit/bc2fbef2bfed4ebc95160cd66f14fe4e23ac8ca2
which landed in master three days ago

benlorenz: Julia will choose whatever version fits the requirements of all packages in the Project.toml.

Having both versions here allows Singular.jl to independently update for CxxWrap.jl 0.10 without us having to worry that it causes unresolvable version requirements for people having both Singular.jl and Polymake.jl in their project.

fingolfin: Well here is the problem with that line: If I install this polymake now, with CxxWrap 0.9.0 installed. It works fine. Then next time I start Julia, this might happen:
```
(v1.3) pkg> update
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
 Installed BinDeps ─ v1.0.1
  Updating `~/.julia/environments/v1.3/Project.toml`
  [1f15a43c] ↑ CxxWrap v0.9.0 ⇒ v0.10.0
  Updating `~/.julia/environments/v1.3/Manifest.toml`
  [9e28174c] ↑ BinDeps v1.0.0 ⇒ v1.0.1
  [1f15a43c] ↑ CxxWrap v0.9.0 ⇒ v0.10.0
  [3eaa8342] + libcxxwrap_julia_jll v0.7.0+0

julia> using Polymake
dyld: lazy symbol binding failed: Symbol not found: __ZN5jlcxx19FunctionWrapperBase19set_pointer_indicesEv
  Referenced from: /Users/mhorn/Projekte/OSCAR/Polymake.jl/src/../deps/src/libpolymake.dylib
  Expected in: flat namespace

dyld: Symbol not found: __ZN5jlcxx19FunctionWrapperBase19set_pointer_indicesEv
  Referenced from: /Users/mhorn/Projekte/OSCAR/Polymake.jl/src/../deps/src/libpolymake.dylib
  Expected in: flat namespace


signal (6): Abort trap: 6
```

Our users will be rightfully upset about that, it's just not an option. I see three possibilities:

1. we stick with 0.9
2. we switch completely to 0.10
3. we come up with a scheme that avoids mixing code compiled against 0.9 with code compiled against 0.10

Regarding option 3, I did something like that in GAP.jl (there, what matters is the Julia version it was compiled against). The solution there was to put all generated code into a directory whose name contains the Julia version (here: the CxxWrap version). This way, the user will still see an error if they upgrade CxxWrap, but if we do things right, that error will be something like "You need to run Pkg.build() first before this work" instead of a crash. Actually, in GAP.jl I simply call Pkg.build(), because why not, that way it'll just work.

benlorenz: This is unfortunate but frequent `Pkg.build` were part of the deal since day one. Julia doesn't even provide a proper way to determine the version of a package (programmatically). (There is `Pkg.installed` but this is marked deprecated without any proper replacement)

The recommended way to avoid this is to build binary dependencies as separate jll packages.

kalmarek: @fingolfin that's a possible scenario, that is true; it's also possible that `libpolymake.so` is linked against `julia-1.3.1` and then `used` in `julia-1.4` (or later) which may or may not be compatible (well, looking at `deps.jl` it's going to be broken anyway);

But the place to fix it is not in `Project.toml` which only states possible compatibility? (at least to my understanding) You don't limit compatibility of GAP to just a single julia version?

fingolfin: We cannot tell referees "oh, you got a crash? well of course you have to run Pkg.build manually from time to time manually, that's part of the deal". We have to work around this day-one deficiency of Julia now, we can't wait for them to clean up their house.

Sure, option 4 might be to use Jlls -- if that works, I am happy to go with it, too! In fact, I've had issues with using your readline builder in GAP.jl (which is not a complaint, it may very well my fault) which are all resolved by switching to Readline_jll;

My point, however, remains: I am opposed to merging this PR in its current form because it results in a situation where a segfault is trivial to trigger and quite likely *will* be triggered (as users who have installed Oscar.jl will have Singular.jl and hence are stuck at CxxWrap 0.9; so Polymake.jl will be compiled against that; once Singular.jl is updated to CxxWrap 0.10, running an update will end up with a segfaulting Polymake.jl; sure the user can fix it but they'll be left with the impression "Oscar crashes" not "Julia is immature" (not that this is much better when presented to our referees).

If we can switch Polymake.jl to Jlls, I'll be happy to switch GAP.jl to them, too; then of course we need to deal with the others -- and all the while make sure Nemo.jl can keep Julia 1.0 compat, which of course adds extra work *sigh* but perhaps we can get away w/o updating Nemo at all; but this must be tested, as we can ill-afford to break Oscar.jl at this point.

So for now, option 1 is maybe more realistic, i.e., we postpone merging this PR?

fingolfin: To clarify; with "this must be tested" I mean: before making a releases of Polymake.jl and GAP.jl which use Jlls, we should verify that installing and running Oscar.jl, Nemo.jl, Singular.jl  still works (e.g. no problems due to multiple copies of GMP). I have reasonably high hopes this will work, but we need manpower to test it all, and if everybody is busy with other things, it might be better to postpone it. But if somebody wants to dive into it, now, be my guest ;-)

kalmarek: but that is yet another issue (almost orthogonal to this one): of how we distribute Oscar;

A solution to this may be to distribute Oscar with `Manifest.toml`. We take control over what and what version is being installed (the official, blessed) and no updates will change that (if I understand correctly how `Manifest.toml` works).

A different one is to provide a binder instance running on our CI server, and instead of asking referees to install something we ask them to visit the following link;

benlorenz:
Lets use the slightly wider comment instead of the review section...

@kalmarek @fingolfin 

I have added a check on the `libcxxwrap-julia` version that was used during build similar to what `CxxWrap.jl` itself does. (I added in the main scope (for precompilation) and in `__init__` to be on the safe side)

This requires everyone that is using Polymake to run `Pkg.build` once to record the current version. (But it will tell you to do so)

Please test!

It should print something like:
```
(Polymake) pkg> up CxxWrap
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Installed BinDeps ─ v1.0.1
   Updating `~/software/polymake/julia/Polymake-glibcxx.jl/Project.toml`
  [1f15a43c] ↑ CxxWrap v0.9.0 ⇒ v0.10.0
   Updating `~/software/polymake/julia/Polymake-glibcxx.jl/Manifest.toml`
  [9e28174c] ↑ BinDeps v1.0.0 ⇒ v1.0.1
  [1f15a43c] ↑ CxxWrap v0.9.0 ⇒ v0.10.0
  [3eaa8342] + libcxxwrap_julia_jll v0.7.0+0

julia> using Polymake
ERROR: LoadError: JlCxx version changed, please run Pkg.build("Polymake");
Stacktrace:
 [1] check_jlcxx_version(::VersionNumber) at /home/lorenz/software/polymake/julia/Polymake-glibcxx.jl/src/Polymake.jl:44
 [2] top-level scope at /home/lorenz/software/polymake/julia/Polymake-glibcxx.jl/src/Polymake.jl:60
 [3] top-level scope at none:2
 [4] eval at ./boot.jl:331 [inlined]
in expression starting at /home/lorenz/software/polymake/julia/Polymake-glibcxx.jl/src/Polymake.jl:60
ERROR: Failed to precompile Polymake [d720cf60-89b5-51f5-aff5-213f193123e7] to /home/lorenz/.julia/compiled/v1.4/Polymake/QizyK_v3eqT.ji.
```

kalmarek:
@benlorenz I confirm this works in both ways (upgrading and downgrading `CxxWrap`)

Project.toml
fingolfin: For now the distribution of `Oscar` is explicitly described as "install the Oscar package in Julia". Of course we can think about doing it differently (we are contemplating this anyway, so that we can produce compiled binaries that avoid most of the startup overhead). But as you say, this is orthogonal to this discussion.

What is *not* orthogonal is this PR breaking the currently used method of distribution Oscar, which can happen as described above.

So, again: this PR cannot be merged as-is. You've now effectively suggested a 5th alternative ("revise way we distribute Oscar so that our users won't be hurt by this PR introducing segfaults, only other users of Polymake.jl will suffer from it"), but I personally think approach 1-4 are superior.

I am pretty sure @fieker and @micjoswig are not keen on merging a PR which will cause people who use Oscar via the way described on our website to risk experiencing segfaults (which, as I said, could happen if this PR is merged, then a release is made, then Singular.jl is updated and released).

And from an unrelated viewpoint: allowing both CxxWrap 0.9 and 0.10 in compat should mean (as I read the Julia documentation) that one can switch between the CxxWrap versions at will w/o breaking anything. Which is not true, as we established above. So I think it breaks at least the spirit of the (admittedly still very limited and underdeveloped) Julia package system.

fieker:
From my limited experience with end users (an old colleague of mine): Pkg.build("Polymake") will not work since it is lacking the using Pkg first.

As to CxxWrap 0.10: by the looks of it, this is a breaking change for Polymake.jl and thus should be in the next major Polymake.jl release. Since OSCAR can and does restrict Polymake to the 0.3 line, we can accomodate this in 0.4
This, of course, make sense only if we assume that Singular is also profiting from CxxWrap 0.10, which could be included in a major Singular release.
Then, we could do a major Oscar release and, possibly, include a deps/build.jl
triggereing a (re)-build of Polymake and Singular.
That is, if CxxWrap 0.10 is useful... or unavaoidable

fieker:
We fought long and hard to be able to do
Pkg.add("Oscar")
if we want/need to get rid of this, there better be a really good reason for it!

kalmarek:
@fingolfin 
> So I think it breaks at least the spirit of the (admittedly still very limited and underdeveloped) Julia package system.

that system has a perfectly fine solution of avoiding the brakage of Oscar.jl: bumping Polymake.jl to `v0.4.0`; but somehow You objected this in the previous pull; let's reconsider this option, shall we?

Project.toml
benlorenz: Since we intended to merge this together with the "drop <1.3" with a new 0.4 version there would be no breakage in oscar until this version would be accepted explicitly into oscar.
So why are you trying to argue against that version bump but complaining about this change. 

Anyway there is now a solution using approach 3, see below.

fingolfin: A version bump to 0.4 here together with a bump to CxxWrap 0.10 would be fine by me. A version bump to 0.4 on that other PR makes no sense to me. Two things can be fine together but not good in isolation, and vice versa.

Anyway, I am glad you went for approach 3, then there is nothing to argue about

fingolfin:
@kalmarek I objected to 0.4.0 *in the context of that other PR*; of course that doesn't mean I am against Polymake.jl ever reaching 0.4.0.

src/Polymake.jl
```diff
@@ -38,6 +38,13 @@ function Base.showerror(io::IO, ex::PolymakeError)
     print(io, "Exception occured at Polymake side:\n$(ex.msg)")
 end
 
+function check_jlcxx_version(version)
+    current_jlcxx = VersionNumber(unsafe_string(ccall(:cxxwrap_version_string, Cstring, ())))
+    if (version != current_jlcxx)
+        error("""JlCxx version changed, please run `using Pkg; Pkg.build("Polymake");`""")
```
fingolfin, line 44: What's the advantage of telling the user this over just calling `Pkg.build("Polymake")`?

benlorenz: I think it is more friendly to give a reason for such a request. Also it makes it easier to debug problems if anyone complains about such a message.

Project.toml
benlorenz: I have added the 0.4.0 bump here.

kalmarek:
are the tests failing on Xcode 11.4 expected?

benlorenz:
yes, see https://github.com/oscar-system/Oscar.jl/issues/82, I am sill investigating this.

PR #255, opened 2020-04-14
Just drop julia < 1.3

benlorenz:
part of #238


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Polymake"
 uuid = "d720cf60-89b5-51f5-aff5-213f193123e7"
 repo = "https://github.com/oscar-system/Polymake.jl.git"
-version = "0.3.3"
+version = "0.4.0"
```
fingolfin, line 4: I still don't understand: why does dropping compat with Julia 1.0-1.2 require a semver change here? If anything running under these older Julia versions depends on Polymake 0.3, they simply won't see any new 0.3.x release with compat Julia < 1.3 dropped;  and code running under Julia >= 1.3 won't mind. What am I missing?

benlorenz: Maybe we need to create a 0.3.4 at some point that is still compatible with julia 1.0. This would be easy by just branching it off before 0.4 and tagging a new 0.3.4.

fingolfin: OK, but how likely is that?

The downside is that once this is changed, we'll also be forced to update Oscar.jl, otherwise it'll not get Polymake.jl updates. So I'd expect the Polymake.jl developers to then submit a PR to Oscar.jl doing just that.

saschatimme: At least for Julia packages with `version >= 1` the auto merge guidelines require for a narrowing of the supported Julia versions a minor version bump.
Couldn't find this written somewhere, but it is in the code:
https://github.com/JuliaRegistries/RegistryCI.jl/blob/6c577e55fd6701b0f6812d39a32922001581cfdb/src/AutoMerge/guidelines.jl#L93

PR #257, opened 2020-04-15
suppress welcome messages

stevengj:
As was also commented in Nemocas/Nemo.jl#817 and [Julia discourse](https://discourse.julialang.org/t/redirecting-stdout-to-avoid-banners-while-import-ing/37633), printing welcome messages upon module import is unusual behavior for Julia modules and quickly becomes unwieldy.   Can this please be omitted?

### DIFF
--- a/src/Polymake.jl
+++ b/src/Polymake.jl
@@ -64,13 +64,8 @@ function __init__()
     end
 
     try
-        show_banner = isinteractive() &&
-                       !any(x->x.name in ["Oscar"], keys(Base.package_locks))
-
-        initialize_polymake(show_banner)
-        if !show_banner
-            shell_execute(raw"$Verbose::credits=\"0\";")
-        end
+        initialize_polymake(false)
+        shell_execute(raw"$Verbose::credits=\"0\";")
     catch ex # initialize_polymake throws jl_error
         throw(PolymakeError(ex.msg))
     end


kalmarek:
I opened the discourse and I'd be very happy to do so! however this is beyond my decision :)
Undoubtely @micjoswig and @fieker would like express their opinions

fieker:
In this, I am pre-empting the main Nemo response: there is a feeling that the GPL requires us to print a banner. IN OSCAR and Hecke we can get away with it, as this requirement is covered by the OSCAR (resp. Hecke) banner.

I'd also add that I think the default Julia "convention" of giving no feedback at all on "using" or "import" is not good at all...it means: hope all worked.

kalmarek:
@fieker no, GPL doesn't require to display anything:
https://www.gnu.org/licenses/gpl-howto.en.html

fieker:
talk to the main Mr Nemo...

fieker:
I know, I've had this discussion several times already. And, while I am goging to defend a banner for Oscar/ Nemo/ Hecke/ Polymake, the GPL is not the reason...
But here's the part that causes anguish:

The startup notice

For interactive programs, it is usually a good idea to make the program display a brief notice about copyright and copying permission when it starts up. See the end of the GNU GPL for more information about this.

...
If the program does terminal interaction, make it output a short notice like this when it starts in an interactive mode:

    <program>  Copyright (C) <year>  <name of author>
    This program comes with ABSOLUTELY NO WARRANTY; for details type `show w'.
    This is free software, and you are welcome to redistribute it
    under certain conditions; type `show c' for details.

stevengj:
First, "usually a good idea" is *advice*, not a legal requirement.

Second, this is **not an interactive "program"**, it is a *library* being loaded by another interactive program (`julia`).

Every GPL *library* that you link to does not display a banner when you run an interactive program.  Otherwise, you'd see zillions of banners for GNU `readline`, GNU `libc`, etcetera.

micjoswig:
> I opened the discourse and I'd be very happy to do so! however this is beyond my decision :)
> Undoubtely @micjoswig and @fieker would like express their opinions

Well, I understand that showing too many banners is undesirable and "unwieldy". On the other hand there is a serious issue with lack of recognition for developers of mathematical software, some of who devote most of their time to writing code.  This is a problem, particularly for young people at the beginning of their (academic) careers.

I agree with viewing the cornerstones ANTIC (Nemo/Hecke), GAP, Polymake and Singular as "libraries" in the greater OSCAR context.  However, I object to viewing OSCAR as a library.  Still in its infancy, this is supposed to be an interactive computer algebra system, making use of Julia.  You could argue that Julia rather plays the "library" role here (despite the possibility of combining OSCAR code with other Julia packages; this is, of course, intended!).

So, I am in favor of producing a joint banner for OSCAR, longer than the current one, with explicit links to the cornerstones at least (as well as FLINT and a few others).  Then, a.f.a.i.a.c the other banners could go.

PR #259, opened 2020-04-17
Add CompatHelper action

fingolfin:
CompatHelper is a Julia package that helps you keep your [compat]
entries up-to-date.

Whenever one of your package's dependencies releases a new version,
CompatHelper opens a pull request on your repository that modifies your
[compat] entry to reflect the newly released version.

See <https://github.com/bcbi/CompatHelper.jl>

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,0 +1,16 @@
+name: CompatHelper
+
+on:
+  schedule:
+    - cron: '00 00 * * *'
+
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pkg.add("CompatHelper")
+        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
+      - name: CompatHelper.main()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: julia -e 'using CompatHelper; CompatHelper.main()'


.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,16 @@
+name: CompatHelper
+
+on:
+  schedule:
+    - cron: '00 00 * * *'
+
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
```
kalmarek, line 9: an array as well?

fingolfin: This file was copied from CompatHelper.jl -- it is *exactly* what they suggest one adds. Of course you can change it, but what's the point?

benlorenz: The array was needed in a matrix specification and is not needed here.

benlorenz:
I don't think this is necessary and I don't like running foreign code on the repository (which gives plenty of access permissions). TagBot at least belong to the JuliaRegistries organization.

fingolfin:
Actually the https://github.com/bcbi organization is the official and verified GitHub organization of an academic department at a major U.S. university. In contrast, https://github.com/JuliaRegistries is an unverified account. We "know" it, so we trust it. But e.g. until I suggested it last week, https://github.com/JuliaActions did not exist, and anybody could have registered it, put up a nice icon and link the Julia website, put int the content from https://github.com/julia-actions (which is official... maybe.. ? I don't know, actual), and then insert malicious content.

This is actually a serious issue for the Julia ecosystem, and it seems the Julia devs are aware of it (see also https://discourse.julialang.org/t/37967).

Note that e.g. https://github.com/JuliaLang/PackageCompiler.jl uses CompatHelper, so if it ever becomes malicious (which I consider rather unlikely), there are far more juicy targets out there.

That said, if you don't want this, that's perfectly fine by me and your discussion, just close the PR :-)

fingolfin:
For the record, [CompatHelper.jl](https://github.com/JuliaRegistries/CompatHelper.jl) since July belongs to the `JuliaRegistries` organization, too (it was moved there some months ago, at least in part motivated by my suggestion to do this as a follow up to the discussion here ;-).

Anyway: While I think this is useful, by no means do I mean to force this upon you. If you don't like it, feel free to just close this PR.

PR #260, opened 2020-04-17
Simplify GitHub actions CI matrix

fingolfin:
Instead of adding three macOS jobs, then removing two of them, it seems
conceptually simpler to just ad the remaining one explicitly.


benlorenz:
I really prefer it the other way round because then any newly added version will be tested with Mac OS as well. Also it better represents the way it was written and it is meant: we do want to test macos but explicitly removed some specific cases to reduce the number of tests.

Also matrix entries require an array: `os: [ ubuntu-latest ]` would have been valid

PR #261, opened 2020-04-17
Wip/mongo

alexej-jordan:
basic support for polyDB

here a small example:

```julia
julia> db = Polymake.Polydb.get_db()
Polymake.Polydb.Database(Database(Client(URI("mongodb://polymake:database@db.polymake.org/?authSource=admin&ssl=true")), "polydb"))

julia> collection = Polymake.Polydb.get_collection(db, "Polytopes.Lattice.SmoothReflexive")
Polymake.Polydb.Collection(Collection(Database(Client(URI("mongodb://polymake:database@db.polymake.org/?authSource=admin&ssl=true")), "polydb"), "Polytopes.Lattice.SmoothReflexive"))

julia> query1 = Dict("DIM"=>3, "N_FACETS"=>5)
Dict{String,Int64} with 2 entries:
  "N_FACETS" => 5
  "DIM"      => 3

julia> results1 = Polymake.Polydb.find(collection, query1)
Polymake.Polydb.Cursor(Mongoc.Cursor{Mongoc.Collection}(Collection(Database(Client(URI("mongodb://polymake:database@db.polymake.org/?authSource=admin&ssl=true")), "polydb"), "Polytopes.Lattice.SmoothReflexive"), Ptr{Nothing} @0x0000000002f29980))

julia> for doc in results1
       println(doc["F_VECTOR"])
       end
Any["6", "9", "5"]
Any["6", "9", "5"]
Any["6", "9", "5"]
Any["6", "9", "5"]
```

any ideas for easily resetting the iterator and how to use more complex BSON queries?


benlorenz:
@kalmarek it would be nice if we could support some advanced syntax for querying with other operators than `=`:
```
find(collection, DIM > 3 && N_VERTICES <= 12)
```
Can we do this via expressions or use some macro to parse this?

kalmarek:
The standard way of filtering collections/finding in collections is through
```
findall(f, A)

  Return a vector I of the indices or keys of A where f(A[I]) returns true. If there are no such elements of A, return an
  empty array.
```
or
```
  filter(f, a::AbstractArray)

  Return a copy of a, removing elements for which f is false. The function f is passed one argument.
```

(the first argument is a predicate)

so this syntax could become
```julia
filter(PolymakeQuery(...), polymake_db)
```
or if you have longer function you could use `do` syntax
```julia
filtered_db = filter(polymake_db) do
    ( ... ) #the content of your PolymakeQuery
end
```

of course `PolymakeQuery` could be a macro expression building the predicate expression and then converting it to the json we'd run Polydb magic on;

---
a different option is to support the api of queryverse: https://github.com/queryverse/Query.jl

to be honest I have no experience with databases, so I don't feel competent enough to have an opinion

kalmarek:
@benlorenz why not? we could add new methods to `findall`/`filter`
```
Base.filter(query::Polydb.Query, polydb::Polymake.Polydb.Collection)
Base.findall(query::Polydb.Query, polydb::Polymake.Polydb.Collection)
```
to do what we want to; then the logic behind could be hidden in `Query` construction and the way we access `Polydb.Collection`. Am I missing something?
e.g. 
```
@query DIM >= 3 && N_VERTICES <= 12
```
constructs the appropriate `Polydb.Query` object and `filter` above takes it, constructs the json and dispatches to the appropriate `Polydb.find` method.

----

as for Query.jl: I don't feel like implementing my own poor-mans query language (the `@query` above), when a solution nearby is present.
```julia
db = Polymake.Polydb.get_db()

result = db |> 
    @select("Polytopes.Lattice.SmoothReflexive") |>
    @filter(_.DIM == 3) |>
    @filter(_.N_FACETS == 5) |>
    @map(_.F_VECTOR) |>
    collect
```
looks clean and we don't have to reinvent the wheel. Of course if we can make it work ;)

alexej-jordan:
for my understanding it would optimal to construct a single `BSON` document that already cotains every condition we want, including e.g. `<` or the logical `OR` and then use that to query the server. locally applying filters does not help is very much, probably forcing us to iterate through a complete collection. passing a `BSON` variable as an additional options argument to `Mongoc.find` is also possible, which is the next i am going to investigate, but i think, besides sorting or reducing memory useage by giving us an object containing only the information we need, this might not be too useful for us

benlorenz:
Now it does make sense ... At the beginning I read your suggestions as we should use these functions but what you meant was probably that we should just use that syntax (which means that we still need to do most of the stuff ourselves but we have an API / Syntax to work along).

The syntax of Query.jl looks nice, we need to investigate how we can overload these macros to construct the corresponding BSON dictionary and dispatch it when something like collect is called.

alexej-jordan:
@benlorenz @kalmarek do we want the objects returned by our iterator to still be the BSON document or should it already be parsed so wa have a `Polymake.BigObject`? in the previous commit i used the first idea, in the commit i just pushed it is the latter.

benlorenz:
I would prefer to hide the BSON from the user. We could/should keep some way to access it directly but default to convert. We might at some point store objects in the database that are not polymake objects (or not only polymake objects).

kalmarek:
we defer query language to #281; I'll try to have a look at the code today

src/polydb.jl
```diff
@@ -0,0 +1,284 @@
...
+   opts = Mongoc.BSON("authorizedCollections" => true, "nameOnly" => true)
+   return Mongoc.get_collection_names(db.mdb;options=opts)
+end
+
+# for the set of names obtained by the _get_collection_names(::Database) function
+# returns two arrays containing the names of the meta data collections
+# first one for sections, second one for collections
+function _get_meta_names(names::Array{String, 1})
+   n = length(names)
+   sec_bool = Array{Bool, 1}(undef, n)
```
kalmarek, line 158: do we need `Array{Bool, 1}`, or would `BitVector` be enough? also please specify (for clarity) that this is `Base.Array` and not `Polymake.Array`

src/polydb.jl
```diff
@@ -0,0 +1,284 @@
...
+         sec_bool[i] = false
+         coll_bool[i] = true
+         n_colls += 1
+      else
+         sec_bool[i] = false
+         coll_bool[i] = false
+      end
+      i += 1
+   end
+   secs = Array{String,1}(undef, n_secs)
```
kalmarek, line 178: `Base.Array`

src/polydb.jl
```diff
@@ -0,0 +1,284 @@
...
+# functions helping printing metadata for sections or collections
+function _print_contact(s::String)
+   println(s)
+end
+
+function _print_contact(a::Array)
+   for dict in a
+      str = dict["name"]
+      for key in ["email", "www", "affiliation"]
+         if haskey(dict, key) && dict[key] != ""
```
kalmarek, line 203: `if !isempty(get(dict, key, ""))`

src/polydb.jl
```diff
@@ -0,0 +1,284 @@
...
+            str = string(str, ", ", dict[key])
+         end
+      end
+      println(string("\t\t", str))
```
kalmarek, line 207: we shouldn't directly print from here. Turn this into `_get_contact` and handle `printing` separately

src/polydb.jl
```diff
@@ -0,0 +1,284 @@
...
+   end
+end
+
+# prints information about a specific section and
+# continues to print information about its content
+function _print_section(db::Database, info::Mongoc.BSON, sections::Array{String,1}, collections::Array{String,1})
+   @info string("SECTION: ", join(info["section"], "."))
```
kalmarek, line 214: note that `@info` goes to `Logging.get_logger()`, while `println` goes to `stdout`. If I capture logs to file I won't see anything except `@info` line

again, it'd be better to separate getting 'string' representation from printing/logging

src/polydb.jl
```diff
@@ -0,0 +1,284 @@
...
+function _evaluateIteration(a::Tuple{Mongoc.BSON, Any})
+   return (parse_document(a[1]), a[2])
+end
+
+# default iteration functions returning `Polymake.BigObject`s
+function Base.iterate(cursor::Cursor)
+   return _evaluateIteration(iterate(cursor.mcursor))
+end
+
+function Base.iterate(cursor::Cursor, state::Nothing)
```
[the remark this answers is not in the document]
alexej-jordan: i understand and checked that back. the problem here is that mostly we just use the iteration on the `Mongoc.Cursor` stored inside the `Polymake.Polydb.Cursor`. and these do not seem to go by this convention:
i modified our `Base.iterate(cursor::Cursor, state::Nothing)` method such that it now prints "HERE" when called. the result:

```julia
julia> for doc in results1
       println(doc.CONE_DIM)
       end
4
HERE
4
HERE
4
HERE
4
HERE
```

as within this iteration we only call the iteration on the internal `Mongoc.Cursor`, this means that `iterate(x::Mongoc.Cursor, ::Nothing) != nothing` for most of the steps. i already wondered about that, but did not bother to look deeper into the implementation of Mongoc.

but that also is the reason why i have this `_evaluateIteration` helper

kalmarek:
ok, I just tried this locally and i get:
```julia
julia> a, s = iterate(results1.mcursor)
ERROR: BSONError: domain=15, code=13053, message=No suitable servers found (`serverSelectionTryOnce` set): [TLS handshake failed: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed calling ismaster on 'db.polymake.org:27017']
```
I guess it's a problem with julia not finding the correct certificates store? @benlorenz?

benlorenz:
> I guess it's a problem with julia not finding the correct certificates store? @benlorenz?

Correct, one workaround might be to set SSL_CERT_FILE to a path containing a CA bundle before starting julia (or SSL_CERT_DIR to a directory with all the certificate files). There might be a bundle in `path/to/julia-$version/share/julia/cert.pem`.

I'm still looking for a better workaround.

benlorenz:
Another workaround is to run `]up` once, directly after launching julia, which loads and initializes the correct SSL libraries, before `using` Mongoc or Polymake.

kalmarek:
ok, the second workaround didn't work for me

benlorenz:
maybe the second one only works on gentoo when using the system libraries, not with the julia bundled ssl libraries.

benlorenz:
I added a fix for the certificate issue, please try without setting any environment variables.

It turns out one can set the CA file via the uri and we can use the one from `MozillaCACerts_jll` which has a variable pointing us to the correct path. This is in my opinion way better than hoping that `joinpath(Sys.BINDIR,Base.DATAROOTDIR,"julia","cert.pem")` exists.

src/polydb.jl
```diff
@@ -0,0 +1,301 @@
...
+function get_db()
+   # we explicitly set the cacert file, otherwise we might get connection errors because the certificate cannot be validated
+   client = Mongoc.Client("mongodb://polymake:database@db.polymake.org/?authSource=admin&ssl=true&sslCertificateAuthorityFile=$(cacert)")
+   return Database(client["polydb"])
+end
+
+# returns a Polymake.Polydb.Collection instance with the given name
+# sections and collections in the name are connected with the '.' sign,
+# i.e. names = "Polytopes.Lattice.SmoothReflexive"
+function get_collection(db::Database, name::String)
```
kalmarek, line 38: since the Database is so close to `Mongoc.Database` why not make it into
```julia
Base.getindex(db::Database, name::AbstractString) = Collection{Polymake.BigObject}(db.mdb[name])
```
also are we sure these will be always `BigObjects`?

this should actually dispatch through the next method as 
`Base.getindex(db::Database, name::AbstractString) = db[name, Polymake.BigObject]`

src/polydb.jl
```diff
@@ -0,0 +1,301 @@
...
+   return Collection{Polymake.BigObject}(db.mdb[name])
+end
+
+function get_collection(db::Database, name::String, ::Type{T}) where T <: Union{Polymake.BigObject, Mongoc.BSON}
+   return Collection{T}(db.mdb[name])
+end
+
+# search a collection for documents matching the criteria given by d
+function find(c::Collection{T}, d::Dict=Dict(); opts::Union{Nothing, Dict}=nothing) where T
```
kalmarek, line 47: why not `Mongoc.find` ?

src/polydb.jl
```diff
@@ -0,0 +1,301 @@
...
+end
+
+function Cursor{Mongoc.BSON}(cursor::Cursor{Polymake.BigObject})
+   return Cursor{Mongoc.BSON}(cursor.mcursor)
+end
+
+# returns a Polymake.BigObject from a Mongoc.BSON document
+function parse_document(bson::Mongoc.BSON)
+   str = Mongoc.as_json(bson)
+   return call_function(:common, :deserialize_json_string, str)
```
kalmarek, line 67: i'd rather use `@pm common.deserialize_json_string(str)`

src/polydb.jl
```diff
@@ -0,0 +1,284 @@
...
+end
+
+#Iterator
+
+# help functions to reduce runtime and keep code clean
+function _evaluateIteration(nothing::Nothing)
```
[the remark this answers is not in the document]
kalmarek: did you have a look at how `Mongoc` handles iteration?

It's only cursor that needs non-intuitive treatment:
```julia
function _iterate(cursor::Cursor, state::Nothing=nothing)
    bson_handle_ref = Ref{Ptr{Cvoid}}()
    has_next = mongoc_cursor_next(cursor.handle, bson_handle_ref)

    if has_next
        # The bson document is valid only until the next call to mongoc_cursor_next.
        # So we should return a deepcopy.
        return deepcopy(BSON(bson_handle_ref[], enable_finalizer=false)), nothing
    else
        err_ref = Ref{BSONError}()
        if mongoc_cursor_error(cursor.handle, err_ref)
            throw(err_ref[])
        end

        return nothing
    end
end

Base.iterate(cursor::Cursor, state::Nothing=nothing) = _iterate(cursor, state)

function Base.iterate(coll::Collection)
    cursor = find(coll)
    return iterate(coll, cursor)
end

function Base.iterate(coll::Collection, state::Cursor)
    next = _iterate(state)
    if next == nothing
        return nothing
    else
        doc, _ = next
        return doc, state
    end
end
```
since our `Cursor/Collection` tightly wrap `Mongoc`s, can we copy it here?

src/polydb.jl
```diff
@@ -0,0 +1,301 @@
...
+   return iterate(coll.mcol)
+end
+
+function Base.iterate(coll::Collection{Mongoc.BSON}, state::Mongoc.Cursor)
+   return iterate(coll.mcol, state)
+end
+
+#Info
+
+# prints a list of the fields of a collection
```
kalmarek, line 124: it doesn't print anymore ;)

src/polydb.jl
```diff
@@ -0,0 +1,301 @@
...
+   else
+      return "no information on fields available"
+   end
+end
+
+function _get_fields(d::Dict)
+   res = ""
+   for (key, value) in d
+      if value == 1
+         res = string(res, key, "\n")
```
kalmarek, line 141: what does a typical field look like?
is returning always a string a good idea? a list of filtered `Pairs` would be much better i guess, something like:
`return map(p -> (last(p) == 1 ? first(p)=>nothing : p), d)`

src/polydb.jl
```diff
@@ -0,0 +1,301 @@
...
+# prints a list of the fields of a collection
+
+function fields(coll::Collection)
+   db = coll.mcol.database
+   coll_c = db[string("_collectionInfo.", coll.mcol.name)]
+   info = collect(coll_c)[2]
+   if haskey(info, "fields")
+      return _get_fields(info["fields"])
+   else
+      return "no information on fields available"
```
kalmarek, line 133: returning strings with warnings is never a good idea.

How about _get_fields return a `Vector{Pair{String, String}}` empty if necessary?

src/polydb.jl
```diff
@@ -0,0 +1,301 @@
...
+
+function _get_fields(a::Array)
+   return join(a, "\n")
+end
+
+# prints information about a specific Collection
+# also used for the info(::Database) function
+function info(coll::Collection)
+   db = coll.mcol.database
+   coll_c = db[string("_collectionInfo.", coll.mcol.name)]
```
kalmarek, line 157: all string operations should happen in here (joining, filtering, etc.)

since we're printing something for the user to inspect visually, this function on the julia side should be rather called `Base.show(io::IO, coll::Collection)`
it also takes middle arguments of type MIME, have a look here: https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing-1

src/polydb.jl
benlorenz: The polymake database contains only BigObjects (that might change at some point but probably not so soon).

alexej-jordan:
> I added a fix for the certificate issue, please try without setting any environment variables.

tested it and it worked

src/polydb.jl
alexej-jordan: here we define the `Polymake.Polydb.find` function, calling `Mongoc.find` in the background. for general purposes in context of the database, the user does not need to `using Mongoc`. The only case this does not apply to is handling the `Mongoc.BSON` templated types, forcing the user to `using Mongoc`. But we could work around that by using another parameter instead of the type; access to entries of `Mongoc.BSON` variables received from polymake still works without `using Mongoc` (i guess because the corresponding `Base.index` method is kept)

src/polydb.jl
alexej-jordan: what exactly do you want to copy in the iteration?
calling the `_evaluateIteration` helper not only sorts out the irregular cursor iteration, inside the documents are parsed to `Polymake.BigObject`s, too. iteration over a `Collection{Polymake.BigObject}` thus also already utilizes the canonical `iterate(x, ::Nothing) = nothing` and otherwise calls `_evaluateIteration(a::Tuple{Mongoc.BSON, Any})` (where parsing happens)

src/polydb.jl
alexej-jordan: i now implemented a `print_fields` and a `get_fields` method. i guess, both are useful.

@benlorenz which entry from the meta collections do you recommend using? at the moment i am asking for the field "fields" of the second entry of a `_collectionInfo`, which contains either an array containing the strings of the fields, a dict with the field name string as key and a comment (or 1 if uncommented) as value or nothing at all.

benlorenz: Don't use `[2]` for the info document, there is no fixed order on these documents, you should fetch them via find using the `_id`. The `fields` entry is not really widely used and maybe unstable.

I think a better (but slightly longer) way is:

1. fetch the info document `find(_collectionInfo, {_id: "info.2.1"})`
2. take the value from the `schema` field, usually just `schema.2.1`
3. fetch the corresponding schema: `find(_collectionInfo, {_id: "schema.2.1"})`
4. read the array `schema["schema"]["required"]`, these are fields that should be present in every object, maybe filter out those starting with an underscore

This should work for all collections except `SmoothReflexive`. The schema is more involved there as the list of fields depends on the dimension of the polytopes (and other properties), I'm not sure if there is an easy way to get a list of properties there. Maybe just merging all `schema["schema"]["allOf"][:]["required"]` arrays, or we could probably parse the schema with polymake and extract it from there.

PS: the 2.1 everywhere refers the current database format version.

src/polydb.jl
kalmarek: my question was rather: since the first argument is `Polymake.Polydb.Collection` why do we chose not to add a method to `Mongoc.find` and rather define a separate `Polymake.Polydb.find` function?

Is there a conflict with signatures?

src/polydb.jl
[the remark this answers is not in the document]
kalmarek: this implements iteration over `Polydb.Cursor`: (nothing else is needed)
```julia
function Base.iterate(cursor::Polymake.Polydb.Cursor{Polymake.BigObject}, state::Nothing=nothing)
    next = iterate(cursor.mcursor, state)
    isnothing(next) && return nothing
    return Polymake.Polydb.parse_document(first(next)), nothing # or last(next)
end

Base.iterate(cursor::Polymake.Polydb.Cursor{Mongoc.BSON}, state::Nothing=nothing) = 
    iterate(cursor.mcursor, state)
```
it seems that (a literal copy of their `iterate`s):
```julia
Base.iterate(coll::Polymake.Polydb.Collection{T}) where T = 
    return iterate(coll, Cursor{T}(find(coll)))

function Base.iterate(coll::Polymake.Polydb.Collection, state::Polymake.Polydb.Cursor)
    next = iterate(state, nothing)
    isnothing(next) && return nothing
    doc, _ = next
    return doc, state
end 
```
should work for collections. small unions are pretty well optimized so there is no need to worry about them (if this was your intention for defining `_evaluateIteration`).

src/polydb.jl
alexej-jordan: the idea behind this is that the user does not have to call `using Mongoc` to access core features regarding polyDB. i would consider the `find` method to be significant enough.
of course we could change the method to `Mongoc.find`, but do we gain anything from it?

kalmarek: If You write here `Mongoc.find` You're adding a new overloaded method that sits inside `Polydb` module. It just extends the other `find`, users will have acces to `Mongoc.find` without specifying the parent module if they're `using Polydb: find`

alexej-jordan: i changed it to `Mongoc.find`, but even when `importing Mongoc` instead of `using` in line 5, i encounter the following problem if i try to use it in a session:

```julia
julia> results1 = Mongoc.find(collection, query1)
ERROR: UndefVarError: Mongoc not defined
Stacktrace:
 [1] top-level scope at none:0

julia> results1 = Polymake.Polydb.find(collection, query1)
ERROR: UndefVarError: find not defined
Stacktrace:
 [1] getproperty(::Module, ::Symbol) at ./Base.jl:26
 [2] top-level scope at none:0

julia> using Mongoc

julia> results1 = Mongoc.find(collection, query1)
Polymake.Polydb.Cursor{Polymake.BigObject}(Mongoc.Cursor{Mongoc.Collection}(Collection(Database(Client(URI("mongodb://polymake:database@db.polymake.org/?authSource=admin&ssl=true&sslCertificateAuthorityFile=/home/jordan/.julia/artifacts/59cb7446ab8874d1c1c4df2eee3cba7c76ed4d1d/share/cacert.pem")), "polydb"), "Polytopes.Lattice.SmoothReflexive"), Ptr{Nothing} @0x000000000c9c1f00))
```

what am i missing?

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+function _get_collection_names(db::Database)
+   opts = Mongoc.BSON("authorizedCollections" => true, "nameOnly" => true)
+   return Mongoc.get_collection_names(db.mdb;options=opts)
+end
+
+# returns an array cotaining the names of all collections in the Polydb, excluding meta collections
+function get_collection_names(db::Database)
+   names = _get_collection_names(db)
+   res = Array{String, 1}()
+   sizehint!(res, Int64(floor(length(names)/2)))
```
kalmarek, line 166:
```suggestion
   sizehint!(res, floor(Int, length(names)/2))
```
or just `sizehint!(res, div(length(names), 2))`

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+   for name in names
+      if SubString(name, 1, 1) != "_"
```
kalmarek, line 168:
```suggestion
      if !startswith(name, "_")
```

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+# first one for sections, second one for collections
+function _get_meta_names(names::Array{String, 1})
+   n = length(names)
+   sec_bool = BitArray{1}(undef, n)
+   coll_bool = BitArray{1}(undef, n)
+   n_secs = 0
+   n_colls = 0
+   i = 1
+   for name in names
+      if SubString(name, 1, 2) == "_s"
```
kalmarek, line 186:
```suggestion
      if startswith(name, "_s")
```

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+         sec_bool[i] = true
+         coll_bool[i] = false
+         n_secs += 1
+      elseif SubString(name, 1, 2) == "_c"
```
kalmarek, line 190:
```suggestion
      elseif startswith(name, "_c")
```

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+   end
+   return res
+end
+
+# for the set of names obtained by the _get_collection_names(::Database) function
+# returns two arrays containing the names of the meta data collections
+# first one for sections, second one for collections
+function _get_meta_names(names::Array{String, 1})
+   n = length(names)
+   sec_bool = BitArray{1}(undef, n)
```
kalmarek, line 180: How about starting both `sec_bool` as `falses(n)` and in the loop over names flipping to `true` when necessary?

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+   coll_bool = BitArray{1}(undef, n)
+   n_secs = 0
+   n_colls = 0
+   i = 1
+   for name in names
```
kalmarek, line 185: whenever we need to iterate over something and keep track of the index it's better to use `enumerate`:
```julia
for (i,name) in enumerate(names)
```

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+         sec_bool[i] = false
+         coll_bool[i] = true
+         n_colls += 1
+      else
+         sec_bool[i] = false
+         coll_bool[i] = false
+      end
+      i += 1
+   end
+   secs = Base.Array{String,1}(undef, n_secs)
```
kalmarek, line 200: if we just need to select names when `sec_bool`/`coll_bool` is true we can do
```julia
secs = names[sec_bool]
colls = names[coll_bool]
```

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+# functions helping printing metadata for sections or collections
+function _get_contact(s::String)
+   return s
+end
+
+function _get_contact(a::Array)
+   res = ""
+   for dict in a
+      str = dict["name"]
+      for key in ["email", "www", "affiliation"]
```
kalmarek, line 225:
```julia
str = join(dict[key] for key in ["name", "email", "www", "affiliation"] if !isempty(get(dict, key, "")), ", ")
```
Are we sure that "name" is always non-empty?
(this is just stylistic, if You find the explicit loop version more readable, leave it as it is)

src/polydb.jl
```diff
@@ -0,0 +1,308 @@
...
+end
+
+# initializes printing complete section/collection tree
+function _get_sections(db::Database, sections::Array{String,1}, collections::Array{String,1})
+   res = ""
+   for sec in sections
+      sec_c = db.mdb[sec]
+      info = iterate(sec_c)[1]
+      if length(info["section"]) == 1
+         res = string(res, _get_section(db, info, sections, collections))
```
kalmarek, line 267: probably this is not crucial here, but every time we allocate another string that grows in length.
It is much better to construct all of them (in an array) and then `join` them at the end

src/polydb.jl
kalmarek: You need to `import Mongoc: find`
when you add new methods to a function it's not automatically brought into scope
```julia
julia> methods(Polymake.Polydb.find)
# 9 methods for generic function "find":
[1] find(bucket::Mongoc.Bucket, bson_filter::Mongoc.BSON; options) in Mongoc at /home/kalmar/.julia/packages/Mongoc/OxMdy/src/gridfs.jl:172
[2] find(bucket::Mongoc.Bucket) in Mongoc at /home/kalmar/.julia/packages/Mongoc/OxMdy/src/gridfs.jl:172
[3] find(collection::Mongoc.CollectionSession, bson_filter::Mongoc.BSON; options) in Mongoc at /home/kalmar/.julia/packages/Mongoc/OxMdy/src/session.jl:93
[4] find(collection::Mongoc.CollectionSession) in Mongoc at /home/kalmar/.julia/packages/Mongoc/OxMdy/src/session.jl:93
[5] find(collection::Mongoc.Collection, bson_filter::Mongoc.BSON; options) in Mongoc at /home/kalmar/.julia/packages/Mongoc/OxMdy/src/collection.jl:203
[6] find(collection::Mongoc.Collection) in Mongoc at /home/kalmar/.julia/packages/Mongoc/OxMdy/src/collection.jl:203
[7] find(c::Polymake.Polydb.Collection{T}) where T in Polymake.Polydb at /home/kalmar/.julia/dev/Polymake/src/polydb.jl:43
[8] find(c::Polymake.Polydb.Collection{T}, d::Dict; opts) where T in Polymake.Polydb at /home/kalmar/.julia/dev/Polymake/src/polydb.jl:43
[9] find(c::Polymake.Polydb.Collection{T}, d::Pair...) where T in Polymake.Polydb at /home/kalmar/.julia/dev/Polymake/src/polydb.jl:47
```

src/polydb.jl
alexej-jordan: by now, "name" always is empty. but including an additional check for future compatibility should not affect runtime noticably in this context

alexej-jordan:
i just pushed a commit in which i completely re-built the our meta-info workflow. i tried to make better use of arrays and strings when generating the long info string, and replaced the recursive algorithm by another one which about does the following:
1. sort the collections and sections into a tree-like nesting of dicts and strings
2. going through the tree in a depth-search-like way, generating an already sorted (due to 1) array of strings
3. joining and printing that array
the way this algorithm works makes some already implemented helpers obsolete, so i removed them.

i still need to add a few comments, but otherwise every suggestion is now implemented.

which leaves the following to be done:
a. querying (issue is already open)
b. info level: should not be to hard to implement now, info level can be checked in the same line as the `haskey` checks, but surely we need to add some more of these to cover all neccessary attributes
c. the jupyter notebook guiding through the usage. im already done 95%, but need to adjust some syntax and need help for the following problem: (@benlorenz @kalmarek )

"Objects from the database know where they come from"
trying to access `a._polyDB` on a `Polymake.BigObject a` received via this module causes the followong exception:

```julia
julia> a.polyDB
ERROR: Exception occured at Polymake side:
unknown property Polytope<Rational>::polyDB at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObjectType.pm line 430.
    Polymake::Core::BigObjectType::property(Polymake::Core::BigObjectType=ARRAY(0x8f4f0c8), "polyDB") called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObjectType.pm line 710
    Polymake::Core::BigObjectType::encode_descending_path(Polymake::Core::BigObjectType=ARRAY(0x8f4f0c8), "polyDB") called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObjectType.pm line 753
    Polymake::Core::BigObjectType::encode_read_request(Polymake::Core::BigObjectType=ARRAY(0x8f4f0c8), "polyDB") called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObject.pm line 1551
    Polymake::Core::BigObject::give_pv called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObject.pm line 1568
    Polymake::Core::BigObject::give(Polymake::polytope::Polytope__Rational=ARRAY(0xadf0410), "polyDB") called at -e line 0
    eval {...} called at -e line 0
```

accessing the "_polyDB" field on a `BSON` object received via this module works without any problems.

probably related, but i can not access `a.DIM` for a `Polymake.BigObject a`, probably because it actually is a function returning `CONE_DIM - 1`?

```julia
julia> a.DIM
ERROR: Exception occured at Polymake side:
unknown property Polytope<Rational>::DIM at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObjectType.pm line 430.
    Polymake::Core::BigObjectType::property(Polymake::Core::BigObjectType=ARRAY(0xaddd668), "DIM") called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObjectType.pm line 710
    Polymake::Core::BigObjectType::encode_descending_path(Polymake::Core::BigObjectType=ARRAY(0xaddd668), "DIM") called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObjectType.pm line 753
    Polymake::Core::BigObjectType::encode_read_request(Polymake::Core::BigObjectType=ARRAY(0xaddd668), "DIM") called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObject.pm line 1551
    Polymake::Core::BigObject::give_pv called at /home/jordan/.julia/dev/Polymake/deps/usr/share/polymake/perllib/Polymake/Core/BigObject.pm line 1568
    Polymake::Core::BigObject::give(Polymake::polytope::Polytope__Rational=ARRAY(0xcc55f48), "DIM") called at -e line 0
    eval {...} called at -e line 0

Stacktrace:
 [1] give(::Polymake.BigObjectAllocated, ::String) at /home/jordan/.julia/dev/Polymake/src/perlobj.jl:45
 [2] getproperty(::Polymake.BigObjectAllocated, ::Symbol) at /home/jordan/.julia/dev/Polymake/src/perlobj.jl:59
```

again, doing this with `BSON` works perfectly fine.

benlorenz:
> "Objects from the database know where they come from"
> trying to access `a._polyDB` on a `Polymake.BigObject a` received via this module causes the followong exception:
> 
> ...
> accessing the "_polyDB" field on a `BSON` object received via this module works without any problems.

`_polyDB` is stored as a normal field in the BSON but converted to an attachment for the polymake objects. On the polymake side there is a method `get_attachment` for `BigObjects` but we have not added this to the julia interface yet (and `attach` as well), I will try to do this tomorrow.

> probably related, but i can not access `a.DIM` for a `Polymake.BigObject a`, probably because it actually is a function returning `CONE_DIM - 1`?
> 
> again, doing this with `BSON` works perfectly fine.

`DIM` is stored for convenience in the BSON to allow queries for DIM, but there is no property `DIM` as this would create a conflict since any `Polytope` with `DIM=2` would also be a `Cone` with `DIM=3`. Thus we have `CONE_DIM` as a property and `DIM` as a method that works depending on the object.
In polymake both can be accessed via `$a->DIM` / `$a->CONE_DIM` but in julia the method is converted to a function (and lowercase): 
```
julia> polytope.dim(c)
3
```

So both effects are as expected.

benlorenz:
Attachment support is in #288, but we will probably need to add support for the type which seems to be a plain perl hash:
```
julia> Polymake.get_attachment(x,"_polyDB")
PropertyValue wrapping HASH
Error showing value of type Polymake.PropertyValueAllocated:
ERROR: invalid value for an input string property
Stacktrace:
 [1] to_string(::Polymake.PropertyValueAllocated) at /home/lorenz/.julia/packages/CxxWrap/LfTHV/src/CxxWrap.jl:597
...
```

In this case it can be converted to `Map{String,String}` which is also unmapped but at least can be displayed:
```
julia> attc = @convert_to Map{String,String} att
PropertyValue wrapping pm::Map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >>
{(collection SmoothReflexive) (creation_date 2019-08-02) (section Polytopes.Lattice) (uri http://polymake.org/polytopes/paffenholz/www/fano.html) (version 2.1)}
```

I will think about a good way to map and convert `HASH`, a nested dictionary with arbitrary and mixed types as values (but only primitive types like int or string as keys).

src/polydb.jl
```diff
@@ -181,39 +185,53 @@ function _get_contact(a::Array)
...
+      push!(res, string("\tFields: ", _get_field_string(db[name])))
+   end
    return join(res, "\n")
 end
 
 # prints a sorted list of the sections and collections of the Polydb
 # together with information about each of these, if existent
 # relying on the structure of Polydb
-function info(db::Database)
+function info(db::Database, level::Int64)
```
kalmarek, line 228: given that on L:233 we define `info(db) = info(db,1)` we could just define
```suggestion
function info(db::Database, level=1)
```
Is `Int64` particularly needed? As far as I can see it's only used for comparison. In such case we can even omit the signature, or if You want to strictly type turn it to `Integer`

PR #263, opened 2020-04-20
use official coverage action

kalmarek:
fixes #231

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -56,6 +56,10 @@ jobs:
         run: echo "::set-env name=DEVELOPER_DIR::/Applications/Xcode_${{ matrix.xcode }}.app/Contents/Developer"
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
-      - uses: julia-actions/julia-uploadcodecov@latest
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v1
+        with:
+          file: ./lcov.info
+          flags: unittests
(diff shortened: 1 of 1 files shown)


.github/workflows/runtests.yml
```diff
@@ -56,6 +56,11 @@ jobs:
...
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v1
+        with:
+          file: ./lcov.info
+          flags: unittests
+          name: codecov-umbrella
+          fail_ci_if_error: false
+          token: ${{ secrets.CODECOV_TOKEN }}
```
benlorenz, line 66:
```suggestion
```
can we try without that token? It should not be necessary and we don't have that in our repository settings.

kalmarek:
@benlorenz have you seen the crash on 1.4, ubuntu-latest ?

benlorenz:
No haven't seen that one yet, will be gone with the next retry I guess.... And no way to debug this...
I like that there is no mention of neither polymake nor cxxwrap in the bracktrace.

But coverage seems to work without the token.

Forcing 0.9.1 blocks the branch-protection rules, shall I change them. Using 0.9 also pulls in 0.9.1.

benlorenz:
I'm seeing a reproducible crash during testing with julia 1.4.1 with my gentoo-julia, maybe this is related. (and maybe specific to 1.4.1 which github is also using now)

I will test with the julia binaries and then open a new ticket depending on the results.

kalmarek:
I removed forcing `0.9.1`, thanks for noticing, this was committed by accident;

I'm running the official 1.4.1 locally (no libraries override) and can not reproduce it, unfortunately;

kalmarek:
and for some reason the crash is gone ;) please bless and merge when appropriate

PR #270, opened 2020-04-25
work around XCode 11.4 typeinfo issue

benlorenz:
I will write more details later, lets see what github actions has to say first.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -148,12 +148,12 @@ pm_libraries = chomp(read(`$perl $pm_config --libs`, String))
 pm_cxx = chomp(read(`$perl $pm_config --cc`, String))
 
 jlcxx_cmake_dir = joinpath(CxxWrap.prefix_path(), "lib", "cmake", "JlCxx")
-julia_exec = joinpath(Sys.BINDIR , "julia")
+julia_exec = joinpath(Sys.BINDIR, Base.julia_exename())
 
 xcode_typeinfo_bug = false
 
 if Sys.isapple()
-   # Lets have some fun for xcode 11.4 until
+   # Work around a bug in Xcode 11.4 that causes SIGABRT, at least until
    # https://github.com/llvm/llvm-project/commit/2464d8135e
    # arrives.
    #
@@ -170,7 +170,7 @@ if Sys.isapple()
    run(`$(CMake.cmake) -DJulia_EXECUTABLE=$julia_exec -DJlCxx_DIR=$jlcxx_cmake_dir .`)
    run(`make -j1`)
    libpath = joinpath(@__DIR__, "xcodetypeinfo", "libhello.$dlext")
-   res = run(pipeline(Cmd(`$julia_exec --project -e "using CxxWrap; @wrapmodule(\"$libpath\", :define_module_hello); @initcxx;"`,ignorestatus=true),stdout=devnull,stderr=devnull))
+   res = run(pipeline(Cmd(`$(Base.julia_cmd()) --project -e "using CxxWrap; @wrapmodule(\"$libpath\", :define_module_hello); @initcxx;"`,ignorestatus=true),stdout=devnull,stderr=devnull))
    if res.termsignal == 6
       global xcode_typeinfo_bug = true
       println("Applying Xcode type_info.hash_code() workaround")
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -152,16 +152,16 @@ julia_exec = joinpath(Sys.BINDIR , "julia")
 
 xcode_typeinfo_bug = false
 
-if (Sys.isapple())
+if Sys.isapple()
    # Lets have some fun for xcode 11.4 until
    # https://github.com/llvm/llvm-project/commit/2464d8135e
    # arrives.
    #
-   # We build a jlcxx library that uses std::string which will should
+   # We build a jlcxx library that uses std::string which will
    # abort with a failed assertion has_julia_type<T> if we are building
    # with xcode 11.4 but libcxxwrap-julia was built with an older libc++.
    #
-   # Read the above commit message some details; the effect of merged
+   # Read the above LLVM commit message for some details; the effect of merged
    # vs non-merged type_info is that for the former memory addresses are
    # used as hash_code(), for the latter the type_info.name() string is
    # hashed.
@@ -173,7 +173,7 @@ if (Sys.isapple())
    res = run(pipeline(Cmd(`$julia_exec --project -e "using CxxWrap; @wrapmodule(\"$libpath\", :define_module_hello); @initcxx;"`,ignorestatus=true),stdout=devnull,stderr=devnull))
    if res.termsignal == 6
       global xcode_typeinfo_bug = true
-      println("Applying XCode type_info.hash_code() workaround")
+      println("Applying Xcode type_info.hash_code() workaround")
    end
 end
 
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -3,6 +3,7 @@ using BinaryProvider
 using Base.Filesystem
 import Pkg
 import CMake
+using Libdl
 
 # Parse some basic command-line arguments
 const verbose = "--verbose" in ARGS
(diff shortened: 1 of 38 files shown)


kalmarek:
wow, that's impressive!

benlorenz:
Some background:
C++ has a typeid function that will return a type_info struct per class, this has a hash_code() method which is used by cxxwrap to identify mapped types. The hash_code is supposed to always give the same result per process per class.

This type_info struct is generated at compilation and merged together from all the objects when linking a shared library.
The hash_code method usually is inline which means that the version present at compilation (in libc++ / libstdc++) is also baked into the object.

In a running process there are different ways to handle multiple copies of the typeinfo object coming from different shared libraries. Some architectures  (e.g. itanium) will guarantee that they are merged together and that there is always only one per class, some don't. For ELF objects this is not guaranteed but mostly true (e.g. with `RTLD_GLOBAL`, as used by default in julia).

But of course the C++ library has to be able to deal with both variants and there are basically two ways used to generate that hash code: either just use the address of the (merged) typeinfo struct, or do a hash of the (mangled) name.

Recently in the libc++ library it was reworked how to distinguish between those cases and as a (unintentional?) side effect the default on mac os was changed from address to name-hash and this was included in xcode 11.4.

And a few days ago this was changed back with an explicit preprocessor define and this will hopefully find its way into xcode soon... (see commit message or comments in the code for a reference to the commit).

The BinaryBuilder uses the same libc++ as the older xcode and thus addresses for the hash.
But Xcode 11.4 uses the name-hash and everything built with this has different hash_code methods based on name-hashing compiled into the objects. Then types are not correctly identified.

benlorenz:
And now for the workaround:
Unfortunately it is not directly possible to change the hash-method as it is a fixed `#define` without a `#ifndef` but we can work around that by first including the `__config` file for libc++ and then redefining the macro `_LIBCPP_HAS_MERGED_TYPEINFO_NAMES_DEFAULT` from 0 to 1.
Then addresses are used for the hash code again, but we have to make sure that the redefine happens before anything else from libc++ is included. For this I have added a `polymake_jlcxx.h` header that does the above and then includes `jlcxx.hpp`. I have slightly reduced the number of headers that need jlcxx and put this extra header instead.

Now, there is another tricky issue:
We need to apply this only when really needed. Setting this always for xcode 11.4 would cause the reverse crash with a xcode 11.4 built custom libcxxwrap-julia.
So we now build and load (in a separate julia process) a small cxxwrap library using std::string that will crash in the cases where we need the workaround. We currently just check whether that process terminated with an abort signal (6), we could also grep the output for "Assertion failed".
In this case we add a custom flag to the CXXFLAGS which enables the redefine-hack.

deps/build.jl
```diff
@@ -149,10 +150,41 @@ pm_cxx = chomp(read(`$perl $pm_config --cc`, String))
...
 julia_exec = joinpath(Sys.BINDIR , "julia")
 
+xcode_typeinfo_bug = false
+
+if (Sys.isapple())
+   # Lets have some fun for xcode 11.4 until
+   # https://github.com/llvm/llvm-project/commit/2464d8135e
+   # arrives.
+   #
+   # We build a jlcxx library that uses std::string which will should
```
fingolfin, line 160:
```suggestion
   # We build a jlcxx library that uses std::string which will
```

deps/build.jl
```diff
@@ -149,10 +150,41 @@ pm_cxx = chomp(read(`$perl $pm_config --cc`, String))
 jlcxx_cmake_dir = joinpath(CxxWrap.prefix_path(), "lib", "cmake", "JlCxx")
 julia_exec = joinpath(Sys.BINDIR , "julia")
 
+xcode_typeinfo_bug = false
+
+if (Sys.isapple())
+   # Lets have some fun for xcode 11.4 until
```
fingolfin, line 156: I'd suggest that we instead make it explicit that we are working around a bug here that causes crashes
```suggestion
   # Work around a bug in Xcode 11.4 that causes segfaults, at least until
```

deps/build.jl
```diff
@@ -149,10 +150,41 @@ pm_cxx = chomp(read(`$perl $pm_config --cc`, String))
 jlcxx_cmake_dir = joinpath(CxxWrap.prefix_path(), "lib", "cmake", "JlCxx")
 julia_exec = joinpath(Sys.BINDIR , "julia")
 
+xcode_typeinfo_bug = false
+
+if (Sys.isapple())
```
fingolfin, line 155:
```suggestion
if Sys.isapple()
```

deps/build.jl
```diff
@@ -149,10 +150,41 @@ pm_cxx = chomp(read(`$perl $pm_config --cc`, String))
...
+   # Lets have some fun for xcode 11.4 until
+   # https://github.com/llvm/llvm-project/commit/2464d8135e
+   # arrives.
+   #
+   # We build a jlcxx library that uses std::string which will should
+   # abort with a failed assertion has_julia_type<T> if we are building
+   # with xcode 11.4 but libcxxwrap-julia was built with an older libc++.
+   #
+   # Read the above commit message some details; the effect of merged
```
fingolfin, line 164:
```suggestion
   # Read the above LLVM commit message for some details; the effect of merged
```

deps/build.jl
```diff
@@ -149,10 +150,41 @@ pm_cxx = chomp(read(`$perl $pm_config --cc`, String))
...
+   # vs non-merged type_info is that for the former memory addresses are
+   # used as hash_code(), for the latter the type_info.name() string is
+   # hashed.
+
+   cd(joinpath(@__DIR__, "xcodetypeinfo"))
+   run(`$(CMake.cmake) -DJulia_EXECUTABLE=$julia_exec -DJlCxx_DIR=$jlcxx_cmake_dir .`)
+   run(`make -j1`)
+   libpath = joinpath(@__DIR__, "xcodetypeinfo", "libhello.$dlext")
+   res = run(pipeline(Cmd(`$julia_exec --project -e "using CxxWrap; @wrapmodule(\"$libpath\", :define_module_hello); @initcxx;"`,ignorestatus=true),stdout=devnull,stderr=devnull))
```
fingolfin, line 173: Should this perhaps use `Base.julia_cmd()`?
```suggestion
   res = run(pipeline(Cmd(`$(Base.julia_cmd()) --project -e "using CxxWrap; @wrapmodule(\"$libpath\", :define_module_hello); @initcxx;"`,ignorestatus=true),stdout=devnull,stderr=devnull))
```

deps/build.jl
```diff
@@ -149,10 +150,41 @@ pm_cxx = chomp(read(`$perl $pm_config --cc`, String))
...
+   if res.termsignal == 6
+      global xcode_typeinfo_bug = true
+      println("Applying XCode type_info.hash_code() workaround")
```
fingolfin, line 176: Nitpick :-)
```suggestion
      println("Applying Xcode type_info.hash_code() workaround")
```

deps/src/polymake_jlcxx.h
```diff
@@ -0,0 +1,23 @@
...
+// workaround xcode 11.4 issue until upstream fix arrives in xcode:
+// https://github.com/llvm/llvm-project/commit/2464d8135e
+//
+// first include __config from libc++, then override typeinfo flag
+// to force use of address as hash instead of hashing the string
+
+#if defined(__APPLE__) && defined(FORCE_XCODE_TYPEINFO_MERGED)
+#include <__config>
+#if defined(_LIBCPP_HAS_MERGED_TYPEINFO_NAMES_DEFAULT) && \
+    _LIBCPP_HAS_MERGED_TYPEINFO_NAMES_DEFAULT == 0
```
fingolfin, line 15: Note that if `_LIBCPP_HAS_MERGED_TYPEINFO_NAMES_DEFAULT` is *not* defined, the above check would result in a compiler warning or even error, as it's expand to `#if false && == 0`.

But of course it should always be defined here, and if it is not, something really funky is going on that we don't know how to deal with, so perhaps this isn't so bad...  *shrug*.

deps/src/polymake_jlcxx.h
```diff
@@ -0,0 +1,23 @@
+#ifndef POLYMAKE_JLCXX
+#define POLYMAKE_JLCXX
+
+// This must be the very first include
+
+// workaround xcode 11.4 issue until upstream fix arrives in xcode:
```
fingolfin, line 6: Super nitpick, but it should be either "work around" (verb) or "workaround for" (noun).

fingolfin:
@mohamed-barakat could you perhaps test this, to see if this "works" for you?

Ideally, test it both with your existing artifacts override, and without.

deps/build.jl
benlorenz: applied but with SIGABRT instead of segfault

deps/src/polymake_jlcxx.h
benlorenz: applied

deps/src/polymake_jlcxx.h
benlorenz: The C preprocessor does shortcut and gives no warning (both clang and gcc) when there is a `defined() &&` before the comparison, even with `-Wall -Wundef`. (I built a small test-program)

But I added an `#else #error` just in case something weird is going on.

mohamed-barakat:
Works with artifacts override:

```
~/.julia/dev/Polymake (pr-270)$ cat ~/.julia/artifacts/Overrides.toml
[3eaa8342-bff7-56a5-9981-c04077f7cee7]
libcxxwrap_julia = "/Users/mo/.julia/cxxwrap-workaround/libcxxwrap-julia-install"
```
```julia
~/.julia/dev/Polymake (pr-270)$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.0 (2020-03-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(@v1.4) pkg> build Polymake
   Building CMake ────→ `~/.julia/packages/CMake/ULbyn/deps/build.log`
   Building LoadFlint → `~/.julia/packages/LoadFlint/42zpR/deps/build.log`
   Building Polymake ─→ `~/.julia/dev/Polymake/deps/build.log`

julia> using Polymake
[ Info: Generating module common
[ Info: Generating module ideal
[ Info: Generating module graph
[ Info: Generating module fulton
[ Info: Generating module fan
[ Info: Generating module group
[ Info: Generating module polytope
[ Info: Generating module topaz
[ Info: Generating module tropical
[ Info: Generating module matroid
polymake version 4.0
Copyright (c) 1997-2020
https://polymake.org

**Comment by mohamed-barakat** [comment]
... and works without artifacts overrides
```
cat ~/.julia/artifacts/Overrides.toml
#[3eaa8342-bff7-56a5-9981-c04077f7cee7]
#libcxxwrap_julia = "/Users/mo/.julia/cxxwrap-workaround/libcxxwrap-julia-install"
```
```julia
~/.julia/dev/Polymake (pr-270)$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.0 (2020-03-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(@v1.4) pkg> build Polymake
   Building CMake ────→ `~/.julia/packages/CMake/ULbyn/deps/build.log`
   Building LoadFlint → `~/.julia/packages/LoadFlint/42zpR/deps/build.log`
   Building Polymake ─→ `~/.julia/dev/Polymake/deps/build.log`

julia> using Polymake
[ Info: Generating module common
[ Info: Generating module ideal
[ Info: Generating module graph
[ Info: Generating module fulton
[ Info: Generating module fan
[ Info: Generating module group
[ Info: Generating module polytope
[ Info: Generating module topaz
[ Info: Generating module tropical
[ Info: Generating module matroid
polymake version 4.0
Copyright (c) 1997-2020
https://polymake.org

```

benlorenz:
Thanks for the reviews and for testing!

PR #273, opened 2020-04-29
Bump version to v0.4.1

saschatimme:
I would really like to have a release with the Xcode 11.4 fix. Is there something holding this bacK?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Polymake"
 uuid = "d720cf60-89b5-51f5-aff5-213f193123e7"
 repo = "https://github.com/oscar-system/Polymake.jl.git"
-version = "0.4.0"
+version = "0.4.1"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"


saschatimme:
Is the `Run tests / test (11.4, 0.9, macos-latest, 1.4` crash expected?

benlorenz:
> Is the `Run tests / test (11.4, 0.9, macos-latest, 1.4` crash expected?

Somewhat expected, this is one of those random crashes with the old cxxwrap 0.9 and julia 1.4. It appears from time to time and might go away when you rerun the builds.

saschatimme:
I see, then I will just accept this for now :)

benlorenz:
@kalmarek wanted to do 0.4.1 in about one week as a replacement for the 0.4.0 but I guess we could also do another minor version in between.

saschatimme:
In my opinion we should make patch releases as often as possible. I understand that we want to be conscious about minor releases but this is just an improvement for everybody.

Any thoughts @kalmarek

kalmarek:
I'm fine releasing this now, minor minor versions are for bugfixes anyway

PR #276, opened 2020-05-05
Fix/startup error

kalmarek:
I crammed three small things here:

* proper error catching during polymake init (occasionally handling resulted in another exception thrown, where thrown exception had no `.msg` (e.g. when it was not actually thrown from withing polymake)
* add `Polymake.version` and `Polymake.installTop` constants (why the latter name? wouldn't `install_prefix` be better? @benlorenz )
* a missing distinction between `BaseMatrix` vs `Polymake.Matrix` in `direct_calls`

### DIFF
--- a/src/Polymake.jl
+++ b/src/Polymake.jl
@@ -87,8 +87,9 @@ function __init__()
         if !show_banner
             shell_execute(raw"$Verbose::credits=\"0\";")
         end
-    catch ex # initialize_polymake throws jl_error
-        throw(PolymakeError(ex.msg))
+    catch ex # initialize_polymake may throw jl_error
+        ex isa ErrorException && throw(PolymakeError(ex.msg))
+        rethrow(ex)
     end
 
     application("common")
--- a/src/polymake_direct_calls.jl
+++ b/src/polymake_direct_calls.jl
@@ -23,6 +23,6 @@ for (pm_solve_LP, scalarT, concreteT) in [
     end
 end
 
-solve_LP(inequalities::AbstractMatrix{T}, objective::AbstractVector; sense=max) where T = solve_LP(inequalities, Matrix{T}(undef, 0, 0), objective, sense)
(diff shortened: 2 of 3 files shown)


benlorenz:
I don't think you can run shell_execute in global scope (i.e. before initializing polymake).

InstallTop is not the prefix but something like `/usr/share/polymake`, there is also InstallArch for libraries (architecture dependent stuff), usually `/usr/lib/polymake`.

The fix for the exception is good, I think I have seen some crashes at this location.

kalmarek:
indeed, you can't. I'm never sure what is the order stuff is evaluated during precompilation

benlorenz:
I guess this is a weird download error that we can't do anything about and should ignore (mac os, julia 1.4):
```
2020-05-06T09:49:29.1333810Z [09:49:28] #################################################                         68.2%
2020-05-06T09:49:29.2761530Z [09:49:29] ##################################################                        70.1%
2020-05-06T09:49:29.4100310Z [09:49:29] ###################################################                       72.1%
2020-05-06T09:49:29.5447180Z [09:49:29] #####################################################                     74.1%
2020-05-06T09:49:29.5452030Z [09:49:29] ######################################################                    76.2%
2020-05-06T09:49:29.5467000Z Assertion failed: (timeout != -1), function uv__io_poll, file src/unix/kqueue.c, line 199.
2020-05-06T09:49:29.5477170Z 
2020-05-06T09:49:29.5479370Z signal (6): Abort trap: 6
2020-05-06T09:49:29.5480090Z in expression starting at /Users/runner/runners/2.169.1/work/Polymake.jl/Polymake.jl/deps/build.jl:77
2020-05-06T09:49:30.8357700Z __pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
2020-05-06T09:49:30.8360060Z Allocations: 49267578 (Pool: 49260206; Big: 7372); GC: 52
2020-05-06T09:49:31.4639750Z â”Œ Error: Error building `Polymake`: 
2020-05-06T09:49:31.4641320Z â”” @ Pkg.Operations /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:892
```

PR #279, opened 2020-05-06
Adjust error message when Polymake needs to be re-built

fingolfin:
Doing `Pkg.build("Polymake")` alone will not work in general; e.g. if
the user only installed Polymake indirectly via `Pkg.add("Oscar")`.

### DIFF
--- a/src/Polymake.jl
+++ b/src/Polymake.jl
@@ -55,7 +55,7 @@ deps_dir = joinpath(@__DIR__, "..", "deps")
 
 isfile(joinpath(deps_dir,"jlcxx_version.jl")) &&
     isfile(joinpath(deps_dir,"deps.jl")) ||
-    error("""Please run `using Pkg; Pkg.build("Polymake");`""")
+    error("""Please run `using Pkg; Pkg.add("Polymake"); Pkg.build("Polymake");`""")
 
 include(joinpath(deps_dir,"jlcxx_version.jl"))
 


benlorenz:
I have tried this several times and cannot reproduce it (even when removing my environment and in a new project). As soon as I have run `Pkg.add("Oscar")` I can also run `Pkg.build("Polymake")` (both with julia 1.3 and 1.4).

`Pkg.build` seems to look into the manifest and successfully finds Polymake (only `Pkg.test` will complain):
```
julia> using Pkg
                                                                                                                                                                                         
julia> Pkg.add("Oscar");                                                                                                                                                                 
   Cloning default registries into `~/.julia`
   Cloning registry from "https://github.com/JuliaRegistries/General.git"                                                                                                                
     Added registry `General` to `~/.julia/registries/General`
 Resolving package versions...                                                                                                                                                           
 Installed URIParser ──────────── v0.4.1
.....
julia> Pkg.build("Polymake")
  Building CMake ────? `~/.julia/packages/CMake/ULbyn/deps/build.log`
  Building LoadFlint ? `~/.julia/packages/LoadFlint/42zpR/deps/build.log`
  Building Polymake ─? `~/.julia/packages/Polymake/wYSJS/deps/build.log`
false

julia> Pkg.test("Polymake")
ERROR: The following package names could not be resolved:
 * Polymake (d720cf60-89b5-51f5-aff5-213f193123e7 in manifest but not in project)
Please specify by known `name=uuid`.
```

kalmarek:
@fingolfin please note that `Pkg.add("Polymake")` introduces changes in users environment that he/she might not intend. It isn't the best idea to decide for/suggest the user to do so.

fingolfin:
Interesting. We run into the problem yesterday, but I can't reproduce it now either. Not sure what went wrong there. Anyway, this is of course better.

PR #282, opened 2020-05-08
build.jl: split `run` argument over multiple lines

fingolfin:
That makes it easier to read it, and even more so diffs on it
when there are changes to this command in a commit.


benlorenz:
thanks!

PR #287, opened 2020-05-18
fix wrapper cleanup during build

benlorenz:
Also:
* use `--no-config` during json generation to ignore `polymake_user` dir and wrappers
* don't clean wrappers at startup

(Originally wrappers were supposed to be cleaned during build and during init but the code didn't actually remove them)


benlorenz:
uncomment the rm?

PR #291, opened 2020-06-01
add many const and a few ref

benlorenz:
Unless there is a good reason (e.g. `!`-functions modifying the argument) we should have `const` refs everywhere.


benlorenz:
Slightly confused by the CI-failure on all platforms ... I dont get those locally.
```
2020-06-01T20:40:56.4334753Z   MethodError: push!(::Polymake.SetAllocated{Int64}, ::Int64) is ambiguous. Candidates:
2020-06-01T20:40:56.4336647Z     push!(arg1::Union{CxxWrap.CxxWrapCore.CxxRef{#s17} where #s17<:Polymake.Set{Int64}, Union{CxxWrap.CxxWrapCore.SmartPointer{T2}, T2} where T2<:Polymake.Set{Int64}}, arg2::Union{Integer, Ref{Int64}}) in Polymake at /home/runner/.julia/packages/CxxWrap/LfTHV/src/CxxWrap.jl:597
2020-06-01T20:40:56.4337145Z     push!(s::Polymake.Set{T}, x::Integer) where T in Polymake at /home/runner/work/Polymake.jl/Polymake.jl/src/sets.jl:131
2020-06-01T20:40:56.4337479Z   Possible fix, define
2020-06-01T20:40:56.4337803Z     push!(::T2 where T2<:Polymake.Set{Int64}, ::Integer)
```

benlorenz:
I will redo this later in `libpolymake_julia`

PR #292, opened 2020-06-09
Mongo/query

alexej-jordan:
here is a raw implementation of advanced querying and the corresponding macros. i followed the example @kalmarek wrote in pull request #261:

```julia
julia> db |>
       Polymake.Polydb.@select("Polytopes.Lattice.SmoothReflexive") |>
       Polymake.Polydb.@filter("DIM" == 3) |>
       Polymake.Polydb.@filter("N_VERTICES" == 8) |>
       Polymake.Polydb.@map() |>
       collect
7-element Array{Polymake.BigObject,1}:
 Polymake.BigObjectAllocated(Ptr{Nothing} @0x000000000b190290)
 Polymake.BigObjectAllocated(Ptr{Nothing} @0x0000000004a9a530)
 Polymake.BigObjectAllocated(Ptr{Nothing} @0x000000000b013fa0)
 Polymake.BigObjectAllocated(Ptr{Nothing} @0x000000000b077b50)
 Polymake.BigObjectAllocated(Ptr{Nothing} @0x000000000b4b4f60)
 Polymake.BigObjectAllocated(Ptr{Nothing} @0x000000000b0ad740)
 Polymake.BigObjectAllocated(Ptr{Nothing} @0x000000000af3f8c0)
```

both `@filter` lines can be joined by just separating the two conditions with a comma.

comparison operators also work, but at the moment only the most important are included.

the `@map` macro does noot do a real job at the moment. calling it without arguments will generate a function asking for every field; only selecting specific fields does not work at the moment (@benlorenz i am sure you can help me with this) when trying to convert to a `Polymake.BigObject`. fields named in the "required" information from the metadata does not seem to be enough. is this due to some metadata missing which is important for generating `Polymake.BigObject`s, like the specific kind of the object? if so, which fields are minimally needed to always be able to convert to a `Polymake.BigObject`?

and, besides coding style, is there anything else important missing? or any operations which we should include etc?


kalmarek:
After discussion with @micjoswig let's keep the user interface very minimal for now, and let's not build too much around it. Looking briefly on the code it seems ok to me. I know we're not extending `Query.jl`, but it's ok for now, we will think about it in the future.

The only thing left is maybe to update the notebook to query language?

benlorenz:
We need some basic tests, which is somewhat tricky as we should not depend on the live polymake database.

We could probably run a [github actions services container](https://help.github.com/en/actions/configuring-and-managing-workflows/about-service-containers#about-service-containers) with mongodb and fill it with a dumped database excerpt using e.g. mongorestore.
This means that we need a way to configure the database uri to switch to the local database, maybe with some environment variable.

In the polymake tests we use the following:
```
docker run --network=something -d --rm=true --network-alias=db-test-server -p 27017:27017 -e MONGO_INITDB_ROOT_USERNAME=admin -e MONGO_INITDB_ROOT_PASSWORD=admin mongo:4.2
# then there needs to be some delay here until the database is fully up
mongorestore --host db-test-server -u admin -p admin --port 27017 some_dir/mongo/
```
I can provide the mongodump files depending on the required contents (the queries should be restricted to small subsets of the collections, i.e. low dimension).

benlorenz:
For `BigObjects` the required entries in the json are probably `_type` and `_ns`, maybe `_id` as well.
If `_polyDB` or other attachments are kept then `_info` should be present as well as it tells polymake which fields are attachments instead of properties.

Issue #268, opened 2020-04-23
Run tests for tutorials in examples/*


benlorenz:
Either in this repository or in a separate oscar-system repository (Tests probably via github-actions)

cc: @kalmarek @micjoswig

kalmarek:
yep, I raised this issue when the location of the notebooks was discussed on Thursday meeting; I'm 100% for making sure those work. Do you think `Literate.jl` an option for this?

benlorenz:
`Literate.jl` cannot really do what we need right now, it can create a notebook from a julia file but not the other way round.

But it could be an option if we change the notebooks to `.jl` files, which would bring some advantages:
- `Literate.jl` can convert to jupyter notebooks for publishing them
- `ipynb` is rather bad for version control, `jl` is a lot better
- we can just run the scripts in the github actions
- we could easily convert them to markdown pages instead of relying on github showing the ipynb more or less nicely
disadvantages:
- needs one more step when running them from the source
- cannot be edited from within jupyter

opinions @kalmarek @IvanSpirandelli @saschatimme ?

alexej-jordan:
in my exys, the advantages overweigh the disadvantages. especially now, in context of the coming documentation, the generated markdown files sound promising.

but we probably need to exclude the polydb notebook from the tests.

PR #297, opened 2020-06-17
Added example notebooks

IvanSpirandelli:


kalmarek:
I had a look at those notebooks and they seem to be ok.
I would prefer a separate folder with `Project/Manifest.toml` fixing the environment.

saschatimme:
Not related to this particular PR, but we should also update or delete the other example notebook.
(Or use something like [Literate.jl](https://github.com/fredrikekre/Literate.jl) to also integrate them into CI).

kalmarek:
the other notebook is kept by @micjoswig here:
https://github.com/micjoswig/oscar-notebooks/blob/master/polymake/g-vectors_of_random_simplicial_6-polytopes.ipynb
@micjoswig maybe those notebooks created by @IvanSpirandelli should go there as well?

I like @saschatimme suggestion about using `Literate.jl` though

benlorenz:
I would prefer to have them here in the Polymake.jl repository (or a new repo under oscar-system), and they should be tested automatically, see #268.

IvanSpirandelli:
I added the tomls and created folders for the examples.

PR #300, opened 2020-06-22
Mongo/test

alexej-jordan:
at the time of creation this pull request is for testing the github services container for polydb


benlorenz:
can you try adding an env section, e.g.
```
env:
 - POLYDB_SERVER_URI: "mongdb://..."
```
in the yaml and then use `get(ENV, "POLYDB_SERVER_URI", ""] != "" ` to check whether the tests should be run?

alexej-jordan:
@benlorenz would you mind helping me with the uri? i assume this is the only thing missing to make the tests work at the moment. i also already included the majority of tests which worked for both the live database and my local dump

alexej-jordan:
@kalmarek would you please have a look at this pull request?
i am still going to:
1. add some comments to the code
2. update the jupyter notebook
3. add docstrings

and i have two questions you or @benlorenz could answer me:
1. should i extend the `@map` macro as soon as possible so that we can feed it the name of fields to be fetched (along with the minimal necessary fields) or should i do this later so that we can push the release of this feature first?
2. by now, i do not check the methods printing information. okay, `Base.show(::IO, ::Collection)` should be no problem with `Base.repr`, but what about the other methids like `Polymake.Polydb.info(::Database, ::Integer)`?

benlorenz:
> 1. should i extend the `@map` macro as soon as possible so that we can feed it the name of fields to be fetched (along with the minimal necessary fields) or should i do this later so that we can push the release of this feature first?

I would prefer if we could merge this as soon as possible and add extra stuff later.

> 2. by now, i do not check the methods printing information. okay, `Base.show(::IO, ::Collection)` should be no problem with `Base.repr`, but what about the other methids like `Polymake.Polydb.info(::Database, ::Integer)`?

For now testing these is not really important, and comparing the strings seems fragile. Maybe we can just run the functions but ignore the output, or at least check that is not zero-length. But it would be fine to ignore them for now and add this in a later pull request.

I have one further request: Can you clean up the history a bit (e.g. squash/fixup all those small testing commits together via `git rebase --interactive`) ?

kalmarek:
I'm fine if tests for show/printing/strings are in the form `@test .... isa String`; I will have a look at the latest changes today

benlorenz:
I tried switching the base, which required a merge of master into the other pull request and now there are a bunch of conflicts in polydb.jl here.

kalmarek:
oh my ;/

src/polydb.jl
```diff
@@ -114,9 +116,9 @@ function get_fields(coll::Collection)
    return temp[(!startswith).(temp, "_")]
 end
 
-function _get_field_string(coll::Collection)
-   return join(get_fields(coll), ", ")
-end
+# function _get_field_string(coll::Collection)
+#    return join(get_fields(coll), ", ")
+# end
```
kalmarek, line 121: this should go away

test/runtests.jl
```diff
@@ -10,20 +10,21 @@ end
...
+    # include("arrays.jl")
+    # include("incidencematrix.jl")
+    # include("convert.jl")
+    # include("perlobj.jl")
+    # include("interface_functions.jl")
+    # include("sparsematrix.jl")
+    # include("tropicalnumber.jl")
+    # include("polynomial.jl")
+    # include("pairs.jl")
+    # include("lists.jl")
```
kalmarek, line 28: we want those tests ;)

test/polydb.jl
```diff
@@ -0,0 +1,163 @@
+using Mongoc
+
+@testset "polyDB" begin
+    add_constraints_poly = get(ENV, "POLYDB_SERVER_URI", "") == "" ? ["DIM" => Dict("\$lte" => 3)] : []
```
kalmarek, line 4: what is `lte`?

test/polydb.jl
```diff
@@ -0,0 +1,163 @@
...
+            @test Mongoc.ping(db.mdb.client)["ok"] == 1
+        catch
+            @test "not" == "connected"
+        end
+        collection_bo = db["Polytopes.Lattice.SmoothReflexive"]
+        @test Polymake.Polydb.Collection{Mongoc.BSON}(collection_bo) isa Polymake.Polydb.Collection
+        @test Polymake.Polydb.Collection{Mongoc.BSON}(collection_bo) isa Polymake.Polydb.Collection{Mongoc.BSON}
+        collection_bson = Polymake.Polydb.Collection{Mongoc.BSON}(collection_bo)
+        @test Polymake.Polydb.Collection{Mongoc.BSON}(collection_bo) isa Polymake.Polydb.Collection
+        @test Polymake.Polydb.Collection{Mongoc.BSON}(collection_bo) isa Polymake.Polydb.Collection{Mongoc.BSON}
```
kalmarek, line 28: we're testing the same thing twice here, is it intended?

I guess the first two tests should be with `Polymake.BigObject`, then the second two tests with `Mongoc.BSON` and then define `collection_bson`?

test/polydb.jl
benlorenz: mongo syntax for `less than or equal`, the constraint with which the database dump was created

alexej-jordan:
@benlorenz i am sorry for the conflicts. was it much?

also, when testing the strings, i encountered a problem:
seemingly, only the two collections i run tests on, the meta collections as well as `Tropical.Cubics` are restored with `mongorestore` on my laptop, leading to problems reading the database structure. yesterday i encountered something similar (but not related) when `Polymake.Polydb.get_fields(::Polymake.Polydb.Collection)` suddenly returned something different when testing with the database on my laptop. but in this case, dropping and re-restoring worked.
i will push a commit after this comment to see if this also applies to the github services container.

@kalmarek thank you for reviewing. i adjusted the code accordingly. and for the operations like `$lte` i have created a table `_operationToMongo` which hopefully helps maintaining and expanding the `@filter` macro, if needed.

also, i have added following behaviour (before that, 2 also ran the same tests as 3:
1. github tests on ubuntu-latest run the polydb tests with the github service container
2. all other github tests do not run any polydb test at all
3. tests run by the user access the live database

the problem is, that the tests the user can run are now not covered in the github tests. as we do not want the github tests to rely on the external database, i would suggest to run a single job with only the polydb tests, which accesses the live database and is allowed to fail

benlorenz:
I never tried to resolve the conflicts that github showed, but they seem to have disappeared during the subsequent merge?

I think by default the polydb tests should not run for the users, and we do not run any tests with the live database on github. If someone manually sets POLYDB_SERVER_URI then the tests can run using that database (this might even be the live database).

test/polydb.jl
```diff
@@ -0,0 +1,167 @@
...
+            @test collect(results_bo) isa Array{Polymake.BigObject, 1}
+            @test collect(results_bson) isa Array{Mongoc.BSON, 1}
+        end
+        @testset "Iterator (Collection)" begin
+            @test iterate(collection_bo) isa Tuple{Polymake.BigObject, Polymake.Polydb.Cursor{Polymake.BigObject}}
+            @test iterate(collection_bson) isa Tuple{Mongoc.BSON, Mongoc.Cursor}
+            if (get(ENV, "POLYDB_SERVER_URI", "") != "")
+                @test collect(collection_bo) isa Array{Polymake.BigObject}
+                @test collect(collection_bson) isa Array{Mongoc.BSON}
+            end
```
benlorenz, line 59:
```suggestion
```
We do not want to fetch a full collection.

test/polydb.jl
```diff
@@ -0,0 +1,167 @@
...
+        end
+        @testset "Information" begin
+            @test Polymake.Polydb.get_fields(collection_bo) isa Array{String, 1}
+            fields = Polymake.Polydb.get_fields(collection_bo)
+            @test length(fields) == (get(ENV, "POLYDB_SERVER_URI", "") == "" ? 43 : 44)
```
benlorenz, line 64:
```suggestion
            @test length(fields) > 10
```
this should be safe even for the live database, no need to check for the correct number.

test/polydb.jl
```diff
@@ -0,0 +1,167 @@
...
+            @test fields[1] == "AFFINE_HULL"
```
benlorenz, line 65:
```suggestion
            @test "FACETS" in fields
```
the fields are not really sorted and facets is a better choice as it really describes the objects and should always be present

test/polydb.jl
```diff
@@ -0,0 +1,167 @@
...
+        db = Polymake.Polydb.get_db()
+        collection_bo = db["Polytopes.Lattice.SmoothReflexive"]
+        for (template, access) in   [(Polymake.BigObject, :((x,y) -> getproperty(x, Symbol(y)))),
+                                    (Mongoc.BSON, :((x,y) -> x[y]))]
+            collection = Polymake.Polydb.Collection{template}(collection_bo)
+            @testset "`$template`-templated types" begin
+                if (get(ENV, "POLYDB_SERVER_URI", "") != "")
+                    complete = collect(collection)
+                    @test length(complete) == 25
+                end
```
benlorenz, line 83:
```suggestion
```

benlorenz:
We could also rename `POLYDB_SERVER_URI` to `POLYDB_TEST_URI` to make it more clear that this is used for the tests.

test/polydb.jl
```diff
@@ -0,0 +1,167 @@
+using Mongoc
+
+@testset "polyDB" begin
+    add_constraints_poly = get(ENV, "POLYDB_SERVER_URI", "") == "" ? ["DIM" => Dict("\$lte" => 3)] : []
```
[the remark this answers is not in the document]
alexej-jordan: we can. but if we do not, the tests (as they are) at the database dump will allow us to test if single conditions work before testing with multiple condition input. what is your thought on that?
we could drop the additional constraint if i find single conditions which can only be satisfied if the additional constraint is naturally true (e.g. `N_VERTICES` small enough s.t. `DIM` neccessarily <= 3), but that sounds tedious to adjust.

alexej-jordan:
@benlorenz currently looking into squashing the commits. can i only squash starting at the latest commit, as that is where the head is?

test/polydb.jl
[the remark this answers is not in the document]
benlorenz: I think that only testing with multiple conditions would also be ok.
But if you prefer it that way: `N_VERTICES <= 4` should give objects with `DIM <= 3` only.

benlorenz:
You can only squash at the HEAD, but with a `rebase --interactive <some commit>` (maybe even with `--rebase-merges`) you could rewrite the whole history and for each commit choose whether to keep it, edit it, drop it or squash it into the previous commit. git will just remove all commits and redo them as requested.
Similar to doing it by hand: start a new branch without all those commit and cherry-pick everything that you want and adjusting the commits as needed.

I fear that this might get messy with the history here with several merges involved, but feel free to try it on a separate branch:
```
git checkout -b testing-rebase mongo/test
git rebase --interactive --rebase-merges 20935a0264da327d8ac298344213b1aea3974389~1
```

kalmarek:
what I understand this can be closed as #306 superseded it; please reopen if necessary

PR #306, opened 2020-07-02
Polymake database interface via Mongoc.jl

alexej-jordan:
cleaned up github history of `mongo/test` branch


alexej-jordan:
@benlorenz is the history clean enough this way? is there anything left besides the comments and the docstrings? the first i will do later today, the latter i am going to address in context of the whole documentation, as soon as the more important things are done.

also, the updated version of the jupyter notebook now is included. do we have a way to access the attachments of `Polymake.BigObject`s at the moment? this is the only thing missing for this notebook now

benlorenz:
sorry for the delay. History looks good and a lot cleaner, thanks!

Regarding attachments:
There is a function `get_attachment(BigObject,String)` that can retrieve the `_polyDB` attachment. But it will return a `PropertyValue` containing a plain perl `HASH` that cannot be printed or used right away. One way to access this from julia would be to add a helper function (maybe overloading the database info function) taking a `BigObject`, fetching the attachment and converting this to a `Map{String,String}` (just for this hash in the `_polyDB` attachment, not for any `HASH`). Unfortunately the `Map` type still needs to be wrapped.

```
julia> att = Polymake.get_attachment(obj,"_polyDB");

julia> matt = @pm common.convert_to{Map{String,String}}(att)
PropertyValue wrapping pm::Map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >>
{(collection SmoothReflexive) (creation_date 2019-08-02) (section Polytopes.Lattice) (uri http://polymake.org/polytopes/paffenholz/www/fano.html) (version 2.1)}
```

alexej-jordan:
no problem, thanks for the reply. i will have a look at the attachment wrapping when this pull request is done. the reason i was asking is the "_polyDB" attachment, which i wanted to include in the jupyter notebook. but luckily, this feature is non-essential (but still somewhat important), so we do not have to include it here.

still, i have one big concern: the polyDB dump does not seem to contain all the collections, which led to a problem reading the structure, thus throwing an error when using the `Polymake.Polydb.info(::Database)` method. i am currently facing some compatibility issues so i have not been able to determine the exact problem. i will send you an email about these issues to keep this discussion on topic

benlorenz:
> still, i have one big concern: the polyDB dump does not seem to contain all the collections, which led to a problem reading the structure, thus throwing an error when using the `Polymake.Polydb.info(::Database)` method.

You could adjust `get_collection_names` to use the metadata collections instead of the collections directly, i.e. `if startswith(name, "_collectionInfo")` instead of `if !startswith(name, "_")` and then using a substring in the push.
(The dump cannot contain empty collections, mongodb doesn't support this)

alexej-jordan:
both the old way and your suggestion seem equivalent to me, but none of both worked. so i tested a bit and have been able to find the problem: the "_sectionInfo" meta collections are not contained in the dump, thus causing an error when generating the `String`, because we also include more than the section name. can we just extend the current dump by adding files for these meta collections from another dump or do you have to make a new one?

benlorenz:
The section info is already in the dump:
```
.github/polydb_dump/polydb/_sectionInfo.Manifolds.bson
.github/polydb_dump/polydb/_sectionInfo.Manifolds.metadata.json
.github/polydb_dump/polydb/_sectionInfo.Matroids.bson
.github/polydb_dump/polydb/_sectionInfo.Matroids.metadata.json
.github/polydb_dump/polydb/_sectionInfo.Polytopes.bson
.github/polydb_dump/polydb/_sectionInfo.Polytopes.Combinatorial.bson
.github/polydb_dump/polydb/_sectionInfo.Polytopes.Combinatorial.metadata.json
.github/polydb_dump/polydb/_sectionInfo.Polytopes.Geometric.bson
.github/polydb_dump/polydb/_sectionInfo.Polytopes.Geometric.metadata.json
.github/polydb_dump/polydb/_sectionInfo.Polytopes.Lattice.bson
.github/polydb_dump/polydb/_sectionInfo.Polytopes.Lattice.metadata.json
.github/polydb_dump/polydb/_sectionInfo.Polytopes.metadata.json
.github/polydb_dump/polydb/_sectionInfo.Tropical.bson
.github/polydb_dump/polydb/_sectionInfo.Tropical.metadata.json
```
They also appear in the github logs during restore, e.g.:
```
2020-07-03T01:39:41.6813915Z 2020-07-03T01:39:41.681+0000	reading metadata for polydb._sectionInfo.Polytopes.Combinatorial from .github/polydb_dump/polydb/_sectionInfo.Polytopes.Combinatorial.metadata.json
2020-07-03T01:39:41.6906685Z 2020-07-03T01:39:41.690+0000	restoring polydb._sectionInfo.Polytopes.Combinatorial from .github/polydb_dump/polydb/_sectionInfo.Polytopes.Combinatorial.bson
2020-07-03T01:39:41.6935996Z 2020-07-03T01:39:41.693+0000	no indexes to restore
2020-07-03T01:39:41.6937274Z 2020-07-03T01:39:41.693+0000	finished restoring polydb._sectionInfo.Polytopes.Combinatorial (1 document, 0 failures)
```

benlorenz:
It would be nice if you would provide some more details on what is going wrong...

I am guessing and trying here, on error I hit is:
```
julia> Polymake.Polydb.info(db_test)
ERROR: MethodError: no method matching getindex(::Nothing, ::String)
Stacktrace:
 [1] _get_section_string(::Polymake.Polydb.Database, ::String, ::Int64) at /home/lorenz/software/polymake/julia/Polymake-git.jl/src/polydb.jl:204
 [2] _get_info_strings(::Polymake.Polydb.Database, ::Dict{String,Union{Array{String,1}, Dict}}, ::Int64, ::String) at /home/lorenz/software/polymake/julia/Polymake-git.jl/src/polydb.jl:275
 [3] _get_info_strings at /home/lorenz/software/polymake/julia/Polymake-git.jl/src/polydb.jl:272 [inlined]
 [4] info at /home/lorenz/software/polymake/julia/Polymake-git.jl/src/polydb.jl:244 [inlined]
 [5] info(::Polymake.Polydb.Database) at /home/lorenz/software/polymake/julia/Polymake-git.jl/src/polydb.jl:243
 [6] top-level scope at REPL[8]:1
```

This might be caused by mongodb showing one extra collection `system.views` in the `get_collection_names` when using the local database (because there are no permissions that would restrict this):
```
julia> Polymake.Polydb.get_collection_names(db_test)
17-element Array{String,1}:
 "Polytopes.Lattice.Reflexive"
 "Tropical.TOM"
 "system.views"
 "Polytopes.Lattice.SmoothReflexive"
...
```

But my suggesting using the `_collectionInfo` collections should fix that, it shows the correct 16 collections:
```
julia> Polymake.Polydb.get_collection_names_via_meta(db_test)
16-element Array{String,1}:
 "Polytopes.Combinatorial.CombinatorialTypes"
 "Polytopes.Lattice.Reflexive"
 "Polytopes.Lattice.ExceptionalMaximalHollow"
...
```

And after I changed `_get_db_tree` to use that one, the following also works:
```
julia> Polymake.Polydb.info(db_test)
SECTION: Tropical
	This database contains tropical objects.

	COLLECTION: Tropical.Cubics
	Regular unimodular triangulations of the triple tetrahedron and their motifs.
```

I have pushed that small change to the branch.
On minor thing I noticed while testing that: the comment says that the list of collections is sorted by I didn't see any sort and the output was in a different order when using the live database.

benlorenz:
The test now fails because it doesn't throw a `MethodError` anymore.

alexej-jordan:
interesting. i had dropped and again restored the database on my local harddrive, but i am still not able to access the section meta collections. but obviously, it works here on github, that is enough.
i also added some slight adjustments to the tests and re-phrased the comment which said "sorted". i meant that the collections are sorted into theit (sub-)section, but i guess "structured" is better suited to express this.

at time of this comment most of the test jobs already succeeded, so i would suggest to merge this branch now

benlorenz:
I think this can be merged now, this was already reviewed in the other two pull requests. @kalmarek @saschatimme any objections?

There might be some more feature requests when this is merged and people start using it. But for now this should be more than enough.

benlorenz:
Hm, there is one weird error in nightly now, but this is probably not critical:
```
2020-07-10T07:48:15.9774710Z Information: Error During Test at /home/runner/work/Polymake.jl/Polymake.jl/test/polydb.jl:62
2020-07-10T07:48:15.9775382Z   Test threw exception
2020-07-10T07:48:17.1007311Z   Expression: Polymake.Polydb.get_fields(col) isa Array{String, 1}
2020-07-10T07:48:17.1378951Z   KeyError: key "allOf" not found
2020-07-10T07:48:17.1379071Z   Stacktrace:
2020-07-10T07:48:17.1379169Z     [1] getindex
2020-07-10T07:48:17.1379268Z       @ ./dict.jl:468 [inlined]
2020-07-10T07:48:17.1379528Z     [2] _read_fields(d::Dict{Any,Any})
2020-07-10T07:48:17.1379663Z       @ Polymake.Polydb ~/work/Polymake.jl/Polymake.jl/src/polydb.jl:140
2020-07-10T07:48:17.1379810Z     [3] get_fields(coll::Polymake.Polydb.Collection{Polymake.BigObject})
2020-07-10T07:48:17.1380132Z       @ Polymake.Polydb ~/work/Polymake.jl/Polymake.jl/src/polydb.jl:119
2020-07-10T07:48:17.1380841Z     [4] top-level scope
2020-07-10T07:48:17.1380934Z       @ ~/work/Polymake.jl/Polymake.jl/test/polydb.jl:62
2020-07-10T07:48:17.1381155Z     [5] top-level scope
2020-07-10T07:48:17.1381324Z       @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1115
2020-07-10T07:48:17.1381552Z     [6] top-level scope
2020-07-10T07:48:17.1381658Z       @ ~/work/Polymake.jl/Polymake.jl/test/polydb.jl:61
2020-07-10T07:48:17.1381863Z     [7] top-level scope
2020-07-10T07:48:17.1381977Z       @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1115
2020-07-10T07:48:17.1382197Z     [8] top-level scope
2020-07-10T07:48:17.1382289Z       @ ~/work/Polymake.jl/Polymake.jl/test/polydb.jl:16
2020-07-10T07:48:17.1382497Z     [9] top-level scope
2020-07-10T07:48:17.1382613Z       @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1115
2020-07-10T07:48:17.1382831Z    [10] top-level scope
2020-07-10T07:48:17.1382936Z       @ ~/work/Polymake.jl/Polymake.jl/test/polydb.jl:5
```

alexej-jordan:
i had something similar twice while working on this, but this rather seemed like a random "hiccup" which disappeared without any changes when testing again (just like sometimes the `julia-buildpkg` step fails. i am not sure what causes this, but as i said, it was not really common. still we should keep an eye on this. have you tried running the job again? is there a way to do this without pushing a new commit?

saschatimme:
What is the purpose of the `bson` files in the `.github` folder? Just for CI?

And is there any documentation for all of this? Maybe at least a small section in the README?

benlorenz:
there is a rerun jobs on the action-page (top-right corner).

the bson files contain the database dump for CI (json is metadata, bson is data).

alexej-jordan:
@saschatimme yes, the `.bson` files are used to create the test database for the CI.

i was planning to extend the whole documentation on the corresponding afterwards, including a better documentation. we also have a jupyter notebook in the `examples` folder now. until the documentation is done, we could refer to it in the `README`.

alexej-jordan:
@benlorenz the tests now succeeded for the same commit.

the `get_fields(::Collection)` method should actually be safe, because we only have two ways of reading these, depending on the structure of the meta documents. i.e., usually we read from the `required` field in the meta document, but `Polytopes.Lattice.SmoothReflexive` is the only collection whose fields are read differently. only in this case, we ask for `allOf`.
seeing that the `Mongoc.BSON` object does not have this key (even though it should, as seen in the other jobs), i wonder if this error could have been caused by some network problems.
and, if i remember correctly, when i stumbled over the same issue, it was not `nightly` but `1.4`.

benlorenz:
Maybe just add `haskey` for the `allof` with another `elseif`. And the final else can throw an error and print the problematic dict. Then the code is even safer and we might see what is causing this.

src/polydb.jl
```diff
@@ -0,0 +1,376 @@
...
+# returns a Polymake.Polydb.Database instance
+#
+# the uri of the server can be set in advance by writing its `String` representation
+# into ENV["POLYDB_TEST_URI"]
+# (used to connect to the github services container for testing)
+function get_db()
+   # we explicitly set the cacert file, otherwise we might get connection errors because the certificate cannot be validated
+   client = Mongoc.Client(get(ENV, "POLYDB_TEST_URI", "mongodb://polymake:database@db.polymake.org/?authSource=admin&ssl=true&sslCertificateAuthorityFile=$(cacert)"))
+   return Database(client["polydb"])
+end
```
kalmarek, line 39: if this function is used only in testing infrastructure I suggest to define it in the test suite, or at least rename to
`__get_test_db()`

src/polydb.jl
```diff
@@ -0,0 +1,376 @@
...
+
+Base.iterate(coll::Collection{Mongoc.BSON}) =
+   iterate(coll.mcol)
+
+Base.iterate(coll::Collection{Mongoc.BSON}, state::Mongoc.Cursor) =
+   iterate(coll.mcol, state)
+
+#Info
+
+# returns an `Array{String, 1}` of the fields of a collection
```
kalmarek, line 107: Since we have all of those comments (thanks!), maybe we should turn them into docstrings immediately?

src/polydb.jl
alexej-jordan: actually, this method is used in both normal and testing environment. the `String` passed to the `Mongoc.Client` constructor is determined by whether `ENV["POLYDB_TEST_URI"]` is set. If it is (e.g. in the CI tests), we read the uri from said environment variable. if it is not, we use the uri of the live database

src/polydb.jl
alexej-jordan: yes, i am planning a separate pull for the documentation, which includes docstrings, doctests and automation of generating the documentation within the CI. this will be right after wrapping `Map{String, String}` (which is already done, without the tests), but i think it would be much easier and cleaner if this pull request here was merged into master before.

Issue #314, opened 2020-08-04
Expand functionality of `map` macro (polyDB)


alexej-jordan:
as of now, `@map` can only be called without any arguments to make it work. to enable the user to state what is needed (and thus reduce unneccessary procession of data), i already wrote some code in the already merged pull request for the `polyDB` access. to give an overview of this issue:
 1) the name of the desired fields should be given as the arguments
 2) this list is then extended by the field names which are neccessary for an object of the type contained in the collection to be valid. theoretically we can split this up into three parts:
 a) fields needed for the object to be mathematically valid
 b) fields needed for the object to be computationally valid (i.e. in context of the software)
 c) fields we want to be mandatory, e.g. the `_polyDB` attachment
 3) the field names are converted into the right format for projection and handed to the optional `opts` setting of the `find` method

due to the way `Polymake` works, i assume that 2a and 2b are the same, or are there any exceptions i should be aware of? are these the mandatory fields the `get_fields(::Collection)` method returns in the usual case?
and regarding 2c, i would of course like to know what other fields we want. i assume, only attachments fall under this category, and they all seem important (but the infromation of `_id` is also contained within `_polyDB`). do we always have `_id`, `_polyDB` and `_type`?

benlorenz:
We should preserve `_ns` and `_attrs` as well. These five `_*` elements (`_id`, `_polyDB`, `_type`, `_ns`, `_attrs`) should always be present, otherwise the json-object parsing in polymake might fail.
If the user requests plain BSON instead of a polymake object these could be dropped as well.

Apart from the above we should not try to check / enforce anything, it is the users duty to supply a reasonable list of properties here (as when creating an object directly). Most subsets of properties should define a valid object (it might be a different one or just a big class of objects instead). There may be special cases where some inconsistency / problems can appear, in some of these cases the initial rules of polymake will complain but in other cases this will silently be accepted.

PR #316, opened 2020-08-05
choose fields when calling 'map' macro (polyDB)

alexej-jordan:
the user can now choose which fields are fetched from the database when querying using the macros. `_*` fields are added to the list. works with both `Polymake.BigObject` and `Mongoc.BSON`.

leaving the `_*` fields out for `Mongoc.BSON` appears inconvenient to me because this should probably not happen automatically (and if it should, when? iteration over a `Cursor{Mongoc.BSON}`? querying a `Collection{Mongoc.BSON}`?)

i should probably wait for the `documentation/new` branch to be merged so i can adjust the docstring for this macro, before we can merge this into the master.


alexej-jordan:
macro is now working as intended and docstring is fitting.

this closes #314

benlorenz:
Can we get one test for that feature?

alexej-jordan:
is there already a way to check `Polymake.BigObject`'s property list? the `Polymake.exists(::Polymake.BigObject, ::String)` method might be enough to check, but is leaves us with a trade-off between precision and amount of code in this case.
the `list_properties` method would do, but it has (intentionally?) not been wrapped yet.

benlorenz:
The `list_properties` method has not been wrapped explicitly but is still available via `call_method`:
```julia
julia> c = polytope.cube(3);

julia> x = Polymake.call_method(c,:list_properties)
6
```

Unfortunately our function call interface does not support list context at the moment (only void and scalar), so we cannot access the list but we do get the number of properties which should be sufficient for the test. Maybe in combination with `exists`.
(Perl automatically converts any array to its size when used in scalar context.)

Issue #317, opened 2020-08-08
GitHub Pages build not triggered


alexej-jordan:
The documentation still can not be accessed after merging the branch into master. Looking into what happened during the workflow shows this warning at the end of the deployment process:

> ┌ Warning: Currently the GitHub Pages build is not triggered when using `GITHUB_TOKEN` for authentication. See issue #1177 (https://github.com/JuliaDocs/Documenter.jl/issues/1177) for more information.

the last comment (https://github.com/JuliaDocs/Documenter.jl/issues/1177#issuecomment-640998242) points out that our setup actually should work. most of the commentors (notably before others claimed this solved) suggested to use ssh keys instead, which appears rather inconvenient in this context to me.

benlorenz:
There is `dev` documentation at https://oscar-system.github.io/Polymake.jl/dev/ but no stable docs yet, probably because it needs to run for a release tag for that.

I don't really know about that warning, but docs are deployed in this log: https://github.com/oscar-system/Polymake.jl/runs/955391821?check_suite_focus=true
Does anything else need to happen?

alexej-jordan:
you are most probably right about needing a versioned release before this can work. we should probably add some more badges for the `README` though, e.g. the `dev` documentation.

and yes, within this log, in the `build and deploy` step, line 32, there is this warning. but now it rather looks like an outdated standard warning

kalmarek:
I'm fine with adding ssh key for deploy; The point is that using GITHUB_TOKEN will not trigger other GH actions, hence if we tag no documentation will be build;

I can add those secrets/keys if you wish, but we already had a similar argument about `CompatHelper` #259 which was left hanging for the same reason

benlorenz:
Lets wait for the next tag before we do anything further. According to that ticket it did work in many cases as soon as some manual commit was on the gh-pages branch.

benlorenz:
This can be closed, release and master docs are being built, see https://github.com/oscar-system/Polymake.jl/tree/gh-pages .

PR #319, opened 2020-09-02
workflow: add ssh key secret for tagbot to allow triggering of documenter

benlorenz:
see #317 for previous discussion

dev-docs work fine but no docs were built for 0.5 so I added an ssh key for tagbot to the github secrets (and deploy keys) and added it to the workflow file

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -9,3 +9,4 @@ jobs:
       - uses: JuliaRegistries/TagBot@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          ssh: ${{ secrets.DOCUMENTER_KEY }}
--- a/.github/workflows/createdocumentation.yml
+++ b/.github/workflows/createdocumentation.yml
@@ -6,6 +6,7 @@ on:
       - master
     tags: '*'
   pull_request:
+  workflow_dispatch:
 
 jobs:
   create-documentation:
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Polymake"
 uuid = "d720cf60-89b5-51f5-aff5-213f193123e7"
 repo = "https://github.com/oscar-system/Polymake.jl.git"
-version = "0.5.0"
+version = "0.5.1"
 
 [deps]
 CxxWrap = "1f15a43c-97ca-5a2a-ae31-89f07a497df4"


benlorenz:
What is the best way to generate docs for 0.5 now?

saschatimme:
I think we need to build this locally and push it onto the gh-pages branch manually.

... or we just make a 0.5.1 release 😅

benlorenz:
I have added a `workflow_dispatch` trigger to the `createdocumentation` job to allow running this manually, but this might not work for a branch where this does not exist.

saschatimme (approved):

alexej-jordan:
Documenter ignores additional data at the end of the version tag, so the decoumentation built for e. G. V1.2.3+doc1 will just have the tag v1. 2.3 (https://github.com/oscar-system/Polymake.jl/pull/319).

Looking at the wording within the guise in the link, the versioned releases are assumed to not be triggered by the master, because that is where the dev docs are generated:

"The documentation is deployed as follows:

Documentation built for a tag vX.Y.Z will be stored in a folder vX.Y.Z.

Documentation built from the devbranch branch (master by default) is stored a folder determined by the devurl keyword to deploydocs (dev by default)."

Does documenter assume us to use a different way of building the versioned releases?

benlorenz:
> Documenter ignores additional data at the end of the version tag, so the decoumentation built for e. G. V1.2.3+doc1 will just have the tag v1. 2.3 (#319).

That is convenient, I have tried this for v0.5.0 now.

> Looking at the wording within the guise in the link, the versioned releases are assumed to not be triggered by the master, because that is where the dev docs are generated:
> 
> "The documentation is deployed as follows:
> 
> Documentation built for a tag vX.Y.Z will be stored in a folder vX.Y.Z.
> 
> Documentation built from the devbranch branch (master by default) is stored a folder determined by the devurl keyword to deploydocs (dev by default)."
> 
> Does documenter assume us to use a different way of building the versioned releases?

No I think everything should be fine like it is.

benlorenz:
Stable docs were successfully built by manually pushing a `v0.5.0+docs` tag: https://oscar-system.github.io/Polymake.jl/stable
I can probably remove that tag now.

With the switch to `DOCUMENTER_KEY` I hope that tagbot will also trigger a build next time we register a new version, if not we can do it manually again and try to fix it once more.
(The workflow_dispatch is not really needed now but doesn't hurt.)

Issue #321, opened 2020-09-15
Ctrl-C quitting julia


fieker:
julia> using Polymake
[ Info: Precompiling Polymake [d720cf60-89b5-51f5-aff5-213f193123e7]
[ Info: Generating module common
[ Info: Generating module ideal
[ Info: Generating module graph
[ Info: Generating module fulton
[ Info: Generating module fan
[ Info: Generating module group
[ Info: Generating module polytope
[ Info: Generating module topaz
[ Info: Generating module tropical
[ Info: Generating module matroid
polymake version 4.1
Copyright (c) 1997-2020
Ewgenij Gawrilow, Michael Joswig, and the polymake team
Technische Universität Berlin, Germany
https://polymake.org

This is free software licensed under GPL; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

julia> s = 0; for i=1:100000 global s+= sin(BigFloat(i))*tan(i^2); end
^C
[fieker@pitti src]$ 

ctrl-c is properly caught in the prompt, but when programms are running, it kills julia (the example is any julia command with a non-trivial runtime as I need time to hit ctrl-c after enter)

benlorenz:
We had a patch for this in place but it somehow got lost in the transition to artifacts, I have added it again for the 4.2 bump of `polymake_jll` (https://github.com/JuliaPackaging/Yggdrasil/pull/1695). In the meantime you can apply [this small patch](https://github.com/benlorenz/Yggdrasil/blob/polymake/P/polymake/bundled/patches/sigint.patch) manually to the files in `polymake_jll.artifact_dir`.

fieker:
On Thu, Sep 24, 2020 at 02:10:31AM -0700, Benjamin Lorenz wrote:
> We had a patch for this in place but it somehow got lost in the transition to artifacts, I have added it again for the 4.2 bump of `polymake_jll` (https://github.com/JuliaPackaging/Yggdrasil/pull/1695). In the meantime you can apply [this small patch](https://github.com/benlorenz/Yggdrasil/blob/polymake/P/polymake/bundled/patches/sigint.patch) manually to the files in `polymake_jll.artifact_dir`.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/oscar-system/Polymake.jl/issues/321#issuecomment-698220155

Thanks!
 Claus

[...]

PR #324, opened 2020-10-12
bump libpolymake_julia to 0.2

benlorenz:
and polymake to 4.2


benlorenz:
Tests look ok, nightly fails because of some incompatibility of our `init_block` in `polymake_jll` with the reorganization of `Pkg.Artifacts`, I will try to deal with that soon.

I think we can finally (after many fights with BinaryBuilder) merge this update, any objections @kalmarek @saschatimme ?
(this includes the fix for #321)

kalmarek (approved):

kalmarek:
I'm all for it!

PR #325, opened 2020-10-20
new jll wrappers have PATH as a RefValue instead ...

benlorenz:
the only "public api" would be a do block but that wont work as we need
the path to be set for all polymake code and not just one execution.


kalmarek (approved):

benlorenz:
macos is failing with julia 1.4 because of an issue with the `normaliz_jll` artifact, that will be fixed in Yggdrasil, so I guess we can merge this anyway.

For nightly I will probably need to move the `init_block` from `polymake_jll` to the Polymake.jl init, that will take a bit.

PR #331, opened 2020-11-04
move perl and ninja paths to the beginning of PATH

benlorenz:
Needed to fix https://github.com/oscar-system/LoadFlint.jl/pull/13
(the jll style `perl() do ... end` code also adds all new paths to the beginning but we cannot use that syntax directly)

While I'm there, also remove old-style PATH access, which should not be necessary anymore.

### DIFF
--- a/src/Polymake.jl
+++ b/src/Polymake.jl
@@ -96,8 +96,7 @@ function __init__()
     global user_dir = abspath(joinpath(Pkg.depots1(),"polymake_user"))
     
     # prepare environment variables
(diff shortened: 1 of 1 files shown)


kalmarek (approved):
How about older versions of jll? were they re-generated?

benlorenz:
For most of the jll only the most recent release was updated for the JLLWrappers.

For Perl_jll, there is just one version plus the new rebuild using the same binaries: `[5.30.3+0, 5.30.3+1]`
For Ninja_jll there is `[1.10.0+0, 1.10.1+0]` and only the recent one uses JLLWrappers but there is no version restriction on Ninja_jll anywhere so that should be fine.

PR #332, opened 2020-11-17
polymake_jll: adjust for deps_tree change from init_block to script

benlorenz:
bump versions for dependencies to make sure the new versions are used
and old Polymake.jl versions can keep using the old ones

note that jlpolymake_version_range still allows 0.2 as I forgot to bump
that version in libpolymake_julia 0.3 but that should not matter as the
version is fixed via project.toml as well (and there are no real changes
in the c++ code)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Polymake"
 uuid = "d720cf60-89b5-51f5-aff5-213f193123e7"
 repo = "https://github.com/oscar-system/Polymake.jl.git"
-version = "0.5.2"
+version = "0.5.3"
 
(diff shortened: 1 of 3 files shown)


kalmarek (approved):

benlorenz:
Thanks, I want to wait a bit more and hope that the polydb-nightly test completes because I made one fix for that. But the github actions seem slightly unstable and hanging right now. Might need a restart of the tests if it doesn't finish soon.

benlorenz:
Now (after a restart) there are three jobs that refuse to continue instead of just two.

PR #339, opened 2021-01-28
Fix html for summary in Firefox

saschatimme:
This is only for the Jupyter notebook property description of big 
objects.
Without this fix the little triangle allowing to show/hide property 
values is missing in Firefox since default CSS values are overwritten by 
some CSS loaded in Jupyter (`normalize.css` to be precise).

Before:

Screen Shot 2021-01-27 at 11 01 57

Now:

Screen Shot 2021-01-27 at 11 02 02

### DIFF
--- a/src/visual.jl
+++ b/src/visual.jl
@@ -49,7 +49,12 @@ function Base.show(io::IO, ::MIME"text/html", obj::BigObject)
             key, val = a, ""
         end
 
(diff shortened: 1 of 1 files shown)


kalmarek (approved):

benlorenz:
The errors with 1.3 are caused by the latest cxxwrap, I removed the "required" for those.

PR #342, opened 2021-03-01
CI: test on macos11 / update julia versions

benlorenz:
Let's see if macos 11 is stable enough yet.

Update versions especially to add 1.6, drop a few 1.3 tests, and fix cxxwrap version on the remaining 1.3 tests.


benlorenz:
macos-11 tests seem ok now, but I changed it to just one such test and will not mark it required as the macos-11 runners are still in preview.
Apart from that, this change fixes the failing 1.3 tests by pinning the cxxwrap version.

lkastner (approved):

PR #349, opened 2021-05-25
bump dependencies for polymake 4.4

benlorenz:
the polymake_jll version is now fixed via the libpolymake_julia compat entry
prepare for version 0.5.6

draft until the last libpolymake_julia_jll is registered (i.e. tests for julia 1.6 will probably fail until that is complete)

### DIFF
--- a/.github/workflows/oscar.yml
+++ b/.github/workflows/oscar.yml
@@ -12,6 +12,7 @@ jobs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     env:
       PR_NUMBER: ${{github.event.number}}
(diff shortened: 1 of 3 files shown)


benlorenz:
It seems that the 1.5 binaries also work for 1.6, but for nightly they do not work. Lets see if the 1.6 binaries help.

benlorenz:
- Timeout in Oscar tests was a bug in minkowski_sum_fukuda in polymake which is fixed in polymake_jll 400.400.0+1.
- CxxWrap bug in nightly can be ignored

So I think we can merge this and tag 0.5.6 afterwards.

PR #352, opened 2021-07-26
Replace Array by Vector or Matrix where appropriate

fingolfin:
As discussed during the coding/documentation sprint today, we want to move
towards requiring Julia 1.6 for running doctests. This is easier if we use
Vector/Matrix instead of Array, as Julia prefers printing these type aliases
over printing `Array` in Julia >= 1.6

This patch was created with the following commands:

    git ls-files | xargs perl -pi -e 's/Array\{([^{}]+(\{((?>[^{}]+)|(?2))*\})?),\s*2\s*\}/Matrix{\1}/g'
    git ls-files | xargs perl -pi -e 's/Array\{([^{}]+(\{((?>[^{}]+)|(?2))*\})?),\s*1\s*\}/Vector{\1}/g'
    git ls-files | xargs perl -pi -e 's/Array\{([^{}]+(\{((?>[^{}]+)|(?2))*\})?),\s*1\s*\}/Vector{\1}/g'

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -373,7 +373,7 @@ The following tables explain by example how to quickly translate `polymake` synt
       <code>new Matrix<T></code><br>Container with algebraic operations
     </td>
     <td>
-      <code>Matrix{T} = Array{T, 2}</code><br>**Linear** container with available indexing by <code>2</code>-ples; all algebra attached
+      <code>Matrix{T} = Matrix{T}</code><br>**Linear** container with available indexing by <code>2</code>-ples; all algebra attached
     </td>
   </tr>
   <tr>
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -373,7 +373,7 @@ The following tables explain by example how to quickly translate `polymake` synt
       <code>new Matrix<T></code><br>Container with algebraic operations
     </td>
     <td>
-      <code>Matrix{T} = Matrix{T}</code><br>**Linear** container with available indexing by <code>2</code>-ples; all algebra attached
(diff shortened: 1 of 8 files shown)


docs/src/examples.md
```diff
@@ -373,7 +373,7 @@ The following tables explain by example how to quickly translate `polymake` synt
       <code>new Matrix<T></code><br>Container with algebraic operations
     </td>
     <td>
-      <code>Matrix{T} = Array{T, 2}</code><br>**Linear** container with available indexing by <code>2</code>-ples; all algebra attached
```
fingolfin, line 376: I guess this line should be adjusted... :-)

src/polydb.jl
```diff
@@ -245,7 +245,7 @@ Base.iterate(coll::Collection{Mongoc.BSON}, state::Mongoc.Cursor) =
 
 """
       get_fields(c::Collection)
-Return an `Array{String, 1}` containing the names of the fields of `c`.
+Return an `Vector{String}` containing the names of the fields of `c`.
```
benlorenz, line 248:
```suggestion
Return a `Vector{String}` containing the names of the fields of `c`.
```

src/polydb.jl
```diff
@@ -323,14 +323,14 @@ end
 """
       get_collection_names(db::Database)
 
-Return an `Array{String, 1}` containing the names of all collections in the
+Return an `Vector{String}` containing the names of all collections in the
```
benlorenz, line 326:
```suggestion
Return a `Vector{String}` containing the names of all collections in the
```

benlorenz (commented):
Thanks! Two minor language corrections.

benlorenz (approved):

PR #353, opened 2021-07-26
CI: add julia 1.7 and bump a few julia versions

benlorenz:


benlorenz:
We should probably remove a few versions before merging this, but lets run all those at least once.

PR #354, opened 2021-07-28
prepare for new release using flint 2.8

benlorenz:
They are fixed via `polymake_jll` for a while now and no need to keep updating them here as well. Bump version to prepare for new release supporting FLINT 2.8.


benlorenz:
It seems we need a rebuild of `libpolymake_julia` as well...

PR #356, opened 2021-08-16
Combine incidencematrix

alexej-jordan:
move functionality of `IncidenceMatrix` from Oscar to polymake and access functionality from there. functionality that is logically applicable to an `IncidenceMatrix` should usually be defined on the `Polymake.jl` side.

within Oscar, importing `IncidenceMatrix` from polymake and then exporting it removes the necessity to re-invent the wheel and re-define the required functionality again in Oscar.

### DIFF
--- a/src/incidencematrix.jl
+++ b/src/incidencematrix.jl
@@ -93,7 +93,7 @@ Base.@propagate_inbounds function col(M::IncidenceMatrix, j::Base.Integer)
 end
 
 function _findnz(M::IncidenceMatrix)
-    len = sum(([length(row(M, i)) for i in 1:size(M, 1)]))
+    len = sum(length, (row(M, i) for i in 1:size(M, 1)))
     ri = Base.Vector{Int64}(undef, len)
     ci = Base.Vector{Int64}(undef, len)
     k = 1
@@ -124,34 +124,40 @@ function Base.resize!(M::IncidenceMatrix{Symmetric}, n::Base.Integer)
     _resize!(M,Int64(n),Int64(n))
 end
 
-function Base.show(io::IO, ::MIME"text/plain", M::IncidenceMatrix)
+function Base.show(io::IO, ::MIME{Symbol("text/plain")}, M::IncidenceMatrix)
     m,n = size(M)
+    a,b = displaysize(io)
+    s = min(a - 5, size(M, 1))
     print(io, "$m×$n IncidenceMatrix\n")
-    for i in 1:min(20, size(M, 1))
-        print(io, "($i) - ")
-        join(io, [i for i in M[i,:].s][1:min(20,length(M[i,:].s))], ", ")
-        if (length(M[i,:]) > 20)
+    for i in 1:s
+        t = min(div(b, 2 + ndigits(size(M, 2))) - 1, length(row(M, i)))
+        print(io, "[")
+        join(io, [c for c in row(M, i)][1:t], ", ")
+        if (length(row(M,i)) > t)
             print(io, ", …")
         end
-        print(io, "\n")
+        print(io, "]\n")
     end
-    if (size(M, 1) > 20)
+    if (m > s)
         print(io, "⁝")
     end
 end
 
 function Base.show(io::IOContext, ::MIME{Symbol("text/plain")}, M::IncidenceMatrix)
     m,n = size(M)
+    a,b = displaysize(io)
+    s = min(a - 5, size(M, 1))
     print(io, "$m×$n IncidenceMatrix\n")
-    for i in 1:min(20, size(M, 1))
-        print(io, "($i) - ")
-        join(io, [i for i in M[i,:].s][1:min(20,length(M[i,:].s))], ", ")
-        if (length(M[i,:]) > 20)
+    for i in 1:s
+        t = min(div(b, 2 + ndigits(size(M, 2))) - 1, length(row(M, i)))
+        print(io, "[")
+        join(io, [c for c in row(M, i)][1:t], ", ")
+        if (length(row(M,i)) > t)
             print(io, ", …")
         end
-        print(io, "\n")
+        print(io, "]\n")
     end
-    if (size(M, 1) > 20)
+    if (m > s)
         print(io, "⁝")
     end
 end
--- a/src/incidencematrix.jl
+++ b/src/incidencematrix.jl
@@ -92,6 +92,27 @@ Base.@propagate_inbounds function col(M::IncidenceMatrix, j::Base.Integer)
     return to_one_based_indexing(_col(M, convert(Int64, j)))
 end
 
+function _findnz(M::IncidenceMatrix)
+    len = sum(([length(row(M, i)) for i in 1:size(M, 1)]))
+    ri = Base.Vector{Int64}(undef, len)
+    ci = Base.Vector{Int64}(undef, len)
+    k = 1
+    for i in 1:size(M, 1)
+        for j in row(M, i)
+            ri[k] = i
+            ci[k] = j
+            k += 1
+        end
+    end
+    return (ri, ci)
+end
+
+function SparseArrays.findnz(M::IncidenceMatrix)
+    ri, ci = _findnz(M)
+    len = length(ri)
+    return (ri, ci, trues(len))
+end
+
 function Base.resize!(M::IncidenceMatrix{NonSymmetric}, m::Base.Integer, n::Base.Integer)
     m >= 0 || throw(DomainError(m, "can not resize to a negative length"))
     n >= 0 || throw(DomainError(n, "can not resize to a negative length"))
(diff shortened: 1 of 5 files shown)


src/sparsevector.jl
```diff
@@ -47,3 +47,40 @@ function findnz(vec::SparseVector{T}) where T
     V = to_jl_type(T)[vec[idx] for idx in I]
     return (I, V)
 end
+
+# implementation of SparseVector{Bool} as Set{Integer} with Integer length
```
benlorenz, line 51:
```suggestion
# implementation of SparseVector{Bool} with Int64 length using a polymake Set
```

src/incidencematrix.jl
```diff
@@ -88,3 +123,35 @@ function Base.resize!(M::IncidenceMatrix{Symmetric}, n::Base.Integer)
     n >= 0 || throw(DomainError(n, "can not resize to a negative length"))
     _resize!(M,Int64(n),Int64(n))
 end
+
+function Base.show(io::IO, ::MIME"text/plain", M::IncidenceMatrix)
+    m,n = size(M)
+    print(io, "$m×$n IncidenceMatrix\n")
+    for i in 1:min(20, size(M, 1))
+        print(io, "($i) - ")
+        join(io, [i for i in M[i,:].s][1:min(20,length(M[i,:].s))], ", ")
```
benlorenz, line 132: I am not really convinced of this output format. The `-` can be confused for a sign, so maybe the `(row-index) - ` could be changed to `[row-index] = ` similar to julias sparsevector. Or we can just omit the row-index since all the rows will be present anyway?

Maybe we should print it like SparseMatrixCSC containing `1` (or `true` but this makes for an ugly layout) for the existing entries and `.` otherwise?
```
3×3 SparseMatrixCSC{Int64, Int64} with 2 stored entries:
 ⋅  ⋅  1
 ⋅  2  ⋅
 ⋅  ⋅  ⋅
```

I am also open for other suggestions.

src/sparsevector.jl
```diff
@@ -47,3 +47,40 @@ function findnz(vec::SparseVector{T}) where T
...
+        push!(V.s, n)
+    else
+        delete!(V.s, n)
+    end
+    return val
+end
+
+function Base.show(io::IO, ::MIME"text/plain", V::SparseVectorBool)
+    l = V.l
+    print(io, "Incidence vector with $l entries. `true` for:\n")
```
benlorenz, line 81:
```suggestion
    print(io, "Incidence vector with length $l. `true` for:\n")
```
This might also need to be adjusted so that it matches the output for the IncidenceMatrix.

benlorenz (changes requested):

src/incidencematrix.jl
alexej-jordan: my personal preference was to utilize the special property of an `IncidenceMatrix` that we can conclude every non-zero entry to be `true`, so we just need to state the locations of these entries. this also is close to the `polymake` design. also, i would say the information is easier to to read from this output, because we can instantly determine the indices from reading the print. that is also the reason i included that `row-index` in each line.

of course, the `SparseMatrixCSC` print creates more of a `Matrix` feeling (in that case i would also prefer `1` instead of `true`, julia does so, too, for `BitArray`), but there is no emphasis on the `Incidence`.

src/sparsevector.jl
alexej-jordan: i just pushed a small suggestion for this

src/incidencematrix.jl
```diff
@@ -50,6 +50,20 @@ function IncidenceMatrix{Symmetric}(mat::AbstractSparseMatrix)
     return res
 end
 
+function IncidenceMatrix{NonSymmetric}(TrueIndices::Base.Array{Base.Array{Int64,1},1}) where T
+    m = length(TrueIndices)
+    n = maximum([maximum(set) for set in TrueIndices])
+    res = IncidenceMatrix(m, n)
+    i = 1
+    for set in TrueIndices
```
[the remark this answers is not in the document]
lkastner: What do you think about something like `onesPositions`? Maybe this is also not optimal. One could also come from the Hasse diagram and say something like `edgeRelations`.

lkastner (approved):

src/incidencematrix.jl
```diff
@@ -50,6 +50,20 @@ function IncidenceMatrix{Symmetric}(mat::AbstractSparseMatrix)
     return res
 end
 
+function IncidenceMatrix{NonSymmetric}(TrueIndices::Base.Array{Base.Array{Int64,1},1}) where T
+    m = length(TrueIndices)
+    n = maximum([maximum(set) for set in TrueIndices])
```
kalmarek, line 55: this allocates a vector only to compute its maximum. maybe 
```suggestion
    n = maximum(maximum, TrueIndices)
```
would work?

kalmarek: also I'm not sure if is `Base.Vector{Base.Vector}` strictly necessary here? at `AbstractVector{<:AbstractVector{<:Integer}}` should be just fine?

src/incidencematrix.jl
```diff
@@ -78,6 +92,27 @@ Base.@propagate_inbounds function col(M::IncidenceMatrix, j::Base.Integer)
     return to_one_based_indexing(_col(M, convert(Int64, j)))
 end
 
+function _findnz(M::IncidenceMatrix)
+    len = sum(([length(row(M, i)) for i in 1:size(M, 1)]))
```
kalmarek, line 96: a vector of lengths is allocated here as well. `sum(length, (row(M, i) for i in 1:size(M, 1)))` should do.

Also: hopefully `row(M, i)` does not allocate. The generator here is usually written as `eachrow(M)`, but we probably don't wrap it, do we?

src/incidencematrix.jl
kalmarek: I guess the standard way would be to hook into `SparseMatrixCSC` printing. If we want to see indices directly that's fine, but then it'd be better to use `displaysize(io)` than hardcoding `20` or some other constant.

src/sparsevector.jl
```diff
@@ -47,3 +47,40 @@ function findnz(vec::SparseVector{T}) where T
...
+
+spzeros(::Type{Bool}, n::Base.Integer) = SparseVectorBool(n, Polymake.Set{to_cxx_type(Int64)}())
+
+Base.size(v::SparseVector{Bool}) = (v.l,)
+Base.eltype(::SparseVector{Bool}) = Bool
+
+Base.show(io::IOContext, ::MIME{Symbol("text/plain")}, v::Polymake.SparseVectorBool) = print(io, length(v), "-element SparseVectorBool. `true` for indices:\n", v.s...)
+
+Base.@propagate_inbounds function Base.getindex(V::SparseVector{Bool}, n::Base.Integer)
+    (V.l >= n && n >= 1) || throw(BoundsError(V, n))
```
kalmarek, line 65:
```suggestion
    @boundscheck checkbounds(V, n)
```
(that is assuming that `SparseVector<:AbstractArray`)

src/sparsevector.jl
```diff
@@ -47,3 +47,40 @@ function findnz(vec::SparseVector{T}) where T
...
+    return in(V, n)
+end
+
+Base.@propagate_inbounds function Base.setindex!(V::SparseVector{Bool}, val, n::Base.Integer)
+    (V.l >= n && n >= 1) || throw(BoundsError(V, n))
```
kalmarek, line 70: as above

src/sparsevector.jl
```diff
@@ -47,3 +47,40 @@ function findnz(vec::SparseVector{T}) where T
...
+    if val
```
kalmarek, line 71: note that this will fail if `val` is not `Bool`, so probably it's better to constrain it in the signature

src/incidencematrix.jl
micjoswig: > I am not really convinced of this output format. 

That's really difficult.  I think the standard polymake output, like one set of indices for row (= facet) is the most intuitive and most compact way of printing.  For OSCAR that should be shifted by one, of course.

The data structure, by design, does not prefer rows over columns; in this sense a matrix (like that Julia type you proposed) is more adequate.  Occasionally it is useful to see this matrix (hence the polymake formatting filter `dense`), but most of the time it is too bulky.

src/incidencematrix.jl
```diff
@@ -88,3 +123,41 @@ function Base.resize!(M::IncidenceMatrix{Symmetric}, n::Base.Integer)
...
+        join(io, [c for c in row(M, i)][1:t], ", ")
+        if (length(row(M,i)) > t)
+            print(io, ", …")
+        end
+        print(io, "]\n")
+    end
+    if (m > s)
+        print(io, "⁝")
+    end
+end
```
benlorenz, line 144:
```suggestion
function Base.show(io::IO, tp::MIME{Symbol("text/plain")}, M::IncidenceMatrix) = Base.show(IOContext(io), tp, M)
```
can't we just use the other function here?

src/sparsevector.jl
```diff
@@ -47,3 +47,76 @@ function findnz(vec::SparseVector{T}) where T
...
+function Base.show(io::IO, ::MIME"text/plain", V::SparseVectorBool)
+    t = min(div(displaysize(io)[2], 2 + ndigits(V.l)) - 1, length(V.s))
+    l = V.l
+    print(io, "$l-element SparseVectorBool\n[")
+    join(io, [i for i in V.s][1:t], ", ")
+    if (length(V.s) > t)
+        print(io, ", …")
+    end
+    print(io, "]")
+end
```
benlorenz, line 111:
```suggestion
Base.show(io::IO, tp::MIME"text/plain", V::SparseVectorBool) = Base.show(IOContext(io), tp, V)
```

benlorenz (approved):

PR #369, opened 2021-11-03
included BigObject in Array_suppT

alexej-jordan:
For Oscar issue 772 and general compatibility it is helpful to include `BigObject` in the constant `Array_suppT` so basic functionality for `Array{BigObject}` is accessible.

### DIFF
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -5,7 +5,8 @@ const Array_suppT = Union{Int64, CxxWrap.CxxLong,
                         StdList{StdPair{CxxWrap.CxxLong,CxxWrap.CxxLong}},
                         Set{Int64}, Set{CxxWrap.CxxLong},
                         Array{Int64}, Array{CxxWrap.CxxLong},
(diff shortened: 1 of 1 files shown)


benlorenz:
We need to be very careful with this as the `Array{BigObject}` does not support all the operations that other `Array{SmallObject}` do, and I am not totally sure about the consequences of adding it to this list.

The constructor `Array{BigObject}(Int)` currently does exists but I would prefer not to use it (I tried to remove it at one point but it is in use in other places in Oscar which we should remove at some points). It does create an Array of that size but it will contain invalid objects that might crash the process when printing it.
For this case we might avoid this by adding a special case like this with the type of the BigObject as a first argument to the constructor:
```
function Array{T}(vec::AbstractVector) where T <: BigObject
    @assert length(vec) > 0
    arr = Array{T}(Polymake.bigobject_type(vec[1]), length(vec))
    @inbounds arr .= vec
    return arr
end
```
But this still needs some thought how to deal with an `AbstractVector` of length 0.

PR #373, opened 2021-11-30
Move to scratchspaces and support many more architectures

benlorenz:
New scratchspaces for dependency tree and polymake_user directory, they will be regenerated after each precompilation since the dependencies (artifact directories) might have changed.
The overhead of running this during `__init__` should be very small and it will be skipped once the directory is non-empty.

Metadata files for polymake applications now comes with the `libpolymake_julia_jll` which avoids running polymake during precompilation.

PS: Tests will fail until all the binaries are done.

_Edit:_ The binaries that this PR will use will support many more architectures (e.g. aarch64-apple-darwin)

This requires: https://github.com/oscar-system/libpolymake-julia/pull/28


src/Polymake.jl
```diff
@@ -70,56 +75,41 @@ end
...
-         @warn "Setting `POLYMAKE_CONFIG` to use a custom polymake installation is no longer supported. Please use `Overrides.toml` to override `polymake_jll` and `libpolymake_julia_jll`."
+
+    binpaths = [
+                 @generate_wrappers(lib4ti2_jll),
+                 @generate_wrappers(TOPCOM_jll),
+                 @generate_wrappers(Ninja_jll),
+                 @generate_wrappers(Perl_jll),
+               ]
+    polymake_deps_tree = @get_scratch!("polymake_deps_$(version_key)_$(rand_key)")
+    if (!isdir(joinpath(polymake_deps_tree,"deps")))
```
fingolfin, line 100:
```suggestion
    if !isdir(joinpath(polymake_deps_tree,"deps"))
```

src/Polymake.jl
```diff
@@ -70,56 +75,41 @@ end
...
-const binpaths = [
-                   @generate_wrappers(lib4ti2_jll),
-                   @generate_wrappers(Ninja_jll),
-                   @generate_wrappers(Perl_jll),
-                 ]
-
-const generated_dir = joinpath(@__DIR__, "generated")
+const version_key = "$(VERSION.major).$(VERSION.minor)"
+# unique per precompilation to ensure deps tree is updated
+const rand_key = Random.randstring(10)
```
fingolfin, line 80: Could you elaborate a bit more why this is needed? Can't you just force an update during precompilation? I.e. write something like
```suggestion

regenerate_dep_tree()
```
and that function is also called from `__init__`. That's essentially how I do it in GAP.jl.

The drawback of using a random number is that it leaves a lot of unused scratch spaces around when you end up precompiling this package multiple times.

fingolfin (commented):
Very nice!

src/Polymake.jl
fingolfin: Alternatively I guess in your model it would suffice to delete the scratch space here?

benlorenz: I was trying to follow the advice of not using any scratchspaces during precompilation to avoid baking in any paths that might end up not being writable anymore later on. But since the symlinks in this directory depend on the exact versions of all the involved artifacts I wanted to make sure that different Polymake.jl installations don't interfere with each other.
A slightly less wasteful option would be to use a hash of the versions of all the dependencies (somewhat similar to what the precompilation does), but just using a random key was a lot easier.
Scratchspaces are cleaned up automatically by julia / Pkg.jl.

Since I was able to move the json generation into the jll, these changes might even solve the central installation issue.

> Alternatively I guess in your model it would suffice to delete the scratch space here?

I am not really sure where the tree should be deleted.

benlorenz: Scratch-key now depends on the julia version and source directory, so this should not generate too many directories. And the dependencies tree is regenerated during `__init__` to make sure all artifacts are still in place.

PR #386, opened 2022-01-12
Add support for type of FUNDAMENTAL_GROUP

alexej-jordan:
this required some small overhaul of lists, which previously only contained pairs of integers.

needs changes from branch of same name in `libpolymake_julia` to work.

### DIFF
--- a/src/std/lists.jl
+++ b/src/std/lists.jl
@@ -1,21 +1,23 @@
-Base.eltype(::StdList{StdPair{T, T}}) where T = Pair{T,T}
+Base.eltype(::StdList{StdPair{S, T}}) where {S, T} = Pair{T,T}
 
-function Base.iterate(L::StdList)
+Base.push!(L::StdList{<:StdPair}, a::Pair) = push!(L, StdPair(a))
+
+Base.eltype(::StdList{T}) where T = T
+
+function Base.iterate(L::StdList{T}) where T
     isempty(L) &&  return nothing
     state = beginiterator(L)
     elt = get_element(state)
     increment(state)
-    return Pair(elt), state
+    return elt, state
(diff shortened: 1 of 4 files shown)


src/std/lists.jl
```diff
@@ -1,21 +1,23 @@
-Base.eltype(::StdList{StdPair{T, T}}) where T = Pair{T,T}
+Base.eltype(::StdList{StdPair{S, T}}) where {S, T} = Pair{T,T}
```
kalmarek, line 1:
```suggestion
Base.eltype(::StdList{StdPair{S, T}}) where {S, T} = Pair{S,T}
```

src/std/lists.jl
```diff
@@ -1,21 +1,23 @@
...
 
-function Base.iterate(L::StdList)
+Base.push!(L::StdList{<:StdPair}, a::Pair) = push!(L, StdPair(a))
+
+Base.eltype(::StdList{T}) where T = T
+
+function Base.iterate(L::StdList{T}) where T
```
kalmarek, line 7: why is `T` needed here?

src/std/lists.jl
```diff
@@ -1,21 +1,23 @@
...
     isempty(L) &&  return nothing
     state = beginiterator(L)
     elt = get_element(state)
     increment(state)
-    return Pair(elt), state
+    return elt, state
 end
 
-function Base.iterate(L::StdList, state)
+function Base.iterate(L::StdList{T}, state) where T
```
kalmarek, line 15: see above

kalmarek (changes requested):

benlorenz (approved):

PR #389, opened 2022-02-01
QuadraticExtension and corresponding container types

alexej-jordan:
This needs the changes from the branch of `libypolymake-julia` of the same name.

### DIFF
--- a/src/quadraticextension.jl
+++ b/src/quadraticextension.jl
@@ -1,0 +1,46 @@
+const qe_suppT = Union{Polymake.Rational}
+
+QuadraticExtension{T}(a::Number, b::Number, r::Number) where T<:qe_suppT =
+    QuadraticExtension{T}(convert(T, a), convert(T, b), convert(T, r))
+    
+QuadraticExtension{T}(a::Number) where T<:qe_suppT = QuadraticExtension{T}(a, 0, 0)
+
+QuadraticExtension(x...) = QuadraticExtension{Rational}(x...)
+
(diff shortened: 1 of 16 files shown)


src/quadraticextension.jl
```diff
@@ -0,0 +1,46 @@
...
+Base.:<=(x::QuadraticExtension{T}, y::QuadraticExtension{T}) where T<:qe_suppT = x < y || x == y
+Base.:/(x::QuadraticExtension{T}, y::QuadraticExtension{T}) where T<:qe_suppT = x // y
+
+# no-copy convert
+convert(::Type{<:QuadraticExtension{T}}, qe::QuadraticExtension{T}) where T<:qe_suppT = qe
+
+function convert(to::Type{<:Number}, qe::QuadraticExtension)
+    !iszero(_b(qe)) && !iszero(_r(qe)) && throw(DomainError("Given QuadraticExtension not trivial."))
+    return convert(to, _a(qe))
+end
```
benlorenz, line 40: I don't really like defining the conversion in this way for all number types, there might even be number types that do support these values.
I think we should just define conversions for the relevant rational (and maybe integer) types.

benlorenz (commented):
Looks reasonable, can you add a few tests that checks some properties of a `Polytope{QuadraticExtension}` or will this be done separately? (e.g. using `polytope.dodecahedron()` and maybe defining a new one from some QE-matrices)

benlorenz (approved):

PR #433, opened 2023-04-21
Don't throw an error if `jupyter kernelspec` does not exist

joschmitt:
Catch and ignore an error if the command `jupyter kernelspec` does not exist and redirect the corresponding `stderr` to nowhere.

Background: on my Debian 12 system, the command `jupyter kernelspec` does not exist. This results in a error/help message from jupyter and throws an `InitError` when I try to load Polymake (or Oscar) in a jupyter notebook.

### DIFF
--- a/src/ijulia.jl
+++ b/src/ijulia.jl
@@ -34,24 +34,30 @@ function prepare_jupyter_kernel_for_visualization()
    success = false
 
    # first try with kernelspec command
-   try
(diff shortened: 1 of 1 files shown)


joschmitt:
> PS: I would still recommend installing jupyter-client.

As far as I can see, I have the package `python3-jupyter-client` installed, but the `jupyter kernelspec` command still does not exist. :shrug:

PR #435, opened 2023-05-23
Move test dependencies to `test/Project.toml`

lgoettgens:
See https://github.com/oscar-system/Oscar.jl/issues/2385.


lgoettgens:
The release-CI failures are probably already resolved downstream in https://github.com/oscar-system/Oscar.jl/pull/2171.

PR #437, opened 2023-05-24
Add `Aqua.jl` for some static testing

lgoettgens:
https://github.com/oscar-system/Oscar.jl/issues/2385.

### DIFF
--- a/test/Aqua.jl
+++ b/test/Aqua.jl
@@ -1,0 +1,15 @@
+using Aqua
+
+@testset "Aqua.jl" begin
+   Aqua.test_all(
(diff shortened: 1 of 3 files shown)


lgoettgens:
When reviewing this PR, please check the compat versions introduced here for sensibility.

lgoettgens:
Thanks for your changes!
For `ambiguities`, you could change the `=false` to `=(recursive=false,)` to only consider ambiguities that arise from at least one method contained in `Polymake.jl`. Maybe this then ignores most of the CxxWrap stuff and leaves only some few things to fix.

PR #440, opened 2023-05-30
add OscarNumber type via polymake_oscarnumber extension

benlorenz:
- ~~unit tests still missing~~
- ~~some working examples in the readme~~
- needed a bit of refactoring how the cxxwrap libraries are loaded (each in a separate module) which now requires some automatic import into the main Polymake.jl module

cc: @alexej-jordan 

this draft is also to check whether everything loads on macos as well ....


lkastner (approved):
wow

alexej-jordan (approved):

PR #466, opened 2024-01-01
Fix precompilation warnings downstream

thofma:
Fixes https://github.com/thofma/Hecke.jl/issues/1340 for me (see also https://github.com/JuliaLang/julia/issues/52685).

If you are happy with this, I would also appreciate a quick patch release containing this fix.


thofma:
Thanks!

PR #467, opened 2024-01-23
avoid closure cfunctions

benlorenz:
Maybe fix some errors on M1 with julia nightly.
Still unclear why this only happens with nightly.


fingolfin:
I can reproduce the issue with Julia master on my M1 mac, and also can verify that this PR  fixes `using Polymake` there

lkastner (approved):

PR #520, opened 2025-06-10
adds connection pool for polydb

antonydellavecchia:
I've been experimenting with multiple connections to polydb at the same time and it seems that using a connection pool is recommended.


antonydellavecchia:
this doesn't work well with mullti processing however, and since Polymake isn't thread save yet, there is no need for this.

PR #522, opened 2025-07-14
Set version to 0.12.2

lgoettgens:
To release https://github.com/oscar-system/Polymake.jl/pull/521.

This should fix the unsatisfiable requirements experienced in https://github.com/thofma/Hecke.jl/actions/runs/16265581708/job/45920393879?pr=1915#step:7:57.


lgoettgens:
My motivation here is to disentangle the version bumps of AA/Nemo/Hecke from the Polymake update. And in case that has any complications...
And since there is already a hugh list of compatible versions, I don't see a reason to not add the new ones there as well.

PR #536, opened 2025-09-10
Remove dependency on Pkg

fingolfin:
See https://github.com/JuliaLang/julia/issues/57970 for motivation.

See also https://github.com/thofma/Hecke.jl/pull/1991 and https://github.com/oscar-system/GAP.jl/pull/1179 for related work.


src/Polymake.jl
```diff
@@ -20,7 +20,6 @@ import Base: ==, <, <=, *, -, +, //, ^, div, rem, one, zero,
     union, union!
 import AbstractAlgebra: should_show_banner
 
-import Pkg # needed by `generate_deps_tree.jl` in the `libpolymake-julia` repo
```
fingolfin, line 23: If this comment is still true this PR of course might break things. But I could not see where in `libpolymake-julia` this would matter, so I thought it's easiest to open this PR and then @benlorenz can tell me why it breaks stuff, and then perhaps we can find a way around it :-)

fingolfin: OK, so the bit I missed is that `src/Polymake.jl` in line 163 does
```
include(libpolymake_julia_jll.generate_deps_tree)
```

PR #547, opened 2026-01-07
allow binary wrappers 0.2

benlorenz:


lgoettgens:
Just wanted to do the same thing, but you were faster :)

PR #549, opened 2026-01-25
add cleanup queue for perl based objects

benlorenz:
to make sure they are deleted on the main thread

~~currently with a debug print at process end to check statistics~~

It might be better to move this to the c++ side by overriding the finalizer on the libcxxwrap side (instead of relying on overriding the internal `CxxWrap.CxxWrapCore.delete` function), but for now this does seem to work.

cc: @fingolfin

### DIFF
--- a/src/Polymake.jl
+++ b/src/Polymake.jl
@@ -209,7 +209,7 @@ function __init__()
     mkpath(target(extarch))
 
     for (ext, dirname) in extensions
-       src(name...) = joinpath(ext.artifact_dir, name...)
+       src(name...) = joinpath(ext.find_artifact_dir(), name...)
        force_symlink(src(exttop, dirname), target(exttop, dirname))
        force_symlink(src(extarch, dirname), target(extarch, dirname))
        push!(extensionpaths, target(exttop, dirname))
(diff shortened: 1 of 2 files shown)


src/Polymake.jl
```diff
@@ -307,7 +307,69 @@ function __init__()
     # the data will be cleaned anyway once the iddict is cleared
     Base.atexit() do
         Polymake.oscarnumber_prepare_cleanup()
+        ccall(:jl_safe_printf, Cvoid, (Cstring, ), "cleanup: $(_count_run[]), $(_count_clean[]), $(_count_cleanobj[])\n")
```
benlorenz:
```suggestion
```

src/Polymake.jl
```diff
@@ -307,7 +307,69 @@ function __init__()
...
+# on the finalizer list
+mutable struct _cleanup_helper
+   function _cleanup_helper()
+      finalizer(_perlobj_finalizer, new())
+   end
+end
+
+const _count_run = Ref(0)
+const _count_clean = Ref(0)
+const _count_cleanobj = Ref(0)
```
benlorenz:
```suggestion
```

src/Polymake.jl
```diff
@@ -307,7 +307,69 @@ function __init__()
...
+
+function _perlobj_finalizer(x)
+   global _count_run[] += 1
```
benlorenz:
```suggestion
```

src/Polymake.jl
```diff
@@ -307,7 +307,69 @@ function __init__()
...
+   # only run on main thread
+   if Threads.threadid() != 1
+      # keep self (partially) alive
+      finalizer(_perlobj_finalizer, x)
+      return nothing
+   end
+   global _count_clean[] += 1
+   while !isempty(_pm_cleanup_queue)
+      # clean queued objects
+      global _count_cleanobj[] += 1
```
benlorenz:
```suggestion
```

src/Polymake.jl
```diff
@@ -307,7 +307,69 @@ function __init__()
...
+
+function _perlobj_finalizer(x)
+   global _count_run[] += 1
+   # only run on main thread
+   if Threads.threadid() != 1
+      # keep self (partially) alive
+      finalizer(_perlobj_finalizer, x)
+      return nothing
+   end
+   global _count_clean[] += 1
```
benlorenz:
```suggestion
```

PR #550, opened 2026-01-26
init polymake without userdir

benlorenz:
This avoids some race conditions with user settings and wrappers. There is not much to store anyway, so this should not cause issues. At least that is what I expect, can't really test all relevant user configurations so we will need to ship this and see what happens.

Also use julia tempfile for extension-config.


benlorenz:
Seems to work, this way we cannot extract new required wrapper from CI anymore but this is rarely necessary.

